PR #3, opened 2016-12-29
Use bors for the command, not just bors.toml file

notriddle:


notriddle:
This push failed because it branched off before travis.yml was added. It will work when it merges.

bors r+

PR #81, opened 2017-02-03
Implement "bors try"; a command to run CI without merging it

### LINKED ISSUE #26: Add the ability to try a PR without merging it
This should not go through the Batcher at all. There should be a separate table of trials that are running, because trials do not benefit from being batched or queued (since they're not merging, they can be run with maximum parallelization. It should be triggered by `bors try`. Any text after the word `try` should be copied into the commit message, so that scripts inside the repository can mine this for information on what should be tested; for example, once this is implemented, bors-ng's own Travis CI configuration will do this:

```shell
if echo "$TRAVIS_COMMIT_MESSAGE" | grep 'bors try: ' 2>&1 >/dev/null; do
    test_args=`echo "$TRAVIS_COMMIT_MESSAGE" | sed 's@^bors try: @@'`
else
    test_args=''
fi
mix test $test_args
```

notriddle:
This is merging it with master, but the result goes in a "trying" branch, and master isn't fast-forwarded to do it.

This is contains a bit much copy-and-paste coding, which will be fixed by pulling commonalities out into their own code and processes (some of this is already there, with the common message and state machine templates, but more is needed).

Fixes #26

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -12,7 +12,8 @@ config :aelita2,
 config :aelita2, Aelita2,
   activation_phrase: "bors r+",
   deactivation_phrase: "bors r-",
-  home_url: "https://bors-ng.github.io/"
+  home_url: "https://bors-ng.github.io/",
+  try_phrase: "bors try"
 
 # Configures the endpoint
 config :aelita2, Aelita2.Endpoint,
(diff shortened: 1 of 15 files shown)


notriddle:
bors try

PR #83, opened 2017-02-05
Handle the case where there's a Link header but no next

notriddle:


notriddle:
Why *doesn't* Elixir use `if` to introduce guard clauses? Is there a syntax ambiguity somehow, or do they just prefer it?

notriddle:
This time, I made sure it compiled.

bors r+

PR #105, opened 2017-02-26
Clean up staging.tmp handling

notriddle:
* Put "(ci skip)" in the commit messages
* Instead of copying master to staging.tmp, create an empty commit with "(ci skip)" at that branch (to ensure that staging.tmp is not built because of that commit).
* Delete staging.tmp after starting the build, so that it won't show up in repository viewers (it's just noise)


notriddle:
bors try

notriddle:
Note to readers; [ci skip] actually has square braces, but I don't want this PR to be skipped every time I try to run it. :smile:

PR #151, opened 2017-04-12
Add clarity to `GITHUB_INTEGRATION_PEM`

notriddle:
In the old version, it's confusing whether it wants you to base64-encode the file itself (right)
or copy out the parts of the file that are base64-encoded (wrong).

### DIFF
--- a/app.json
+++ b/app.json
@@ -33,7 +33,7 @@
       "description": "GitHub Integration ID"
     },
     "GITHUB_INTEGRATION_PEM": {
-      "description": "GitHub Integration private key (base64-encoded)"
+      "description": "GitHub Integration private key (download the file, encode it with base64, then paste the result here)"
     },
     "GITHUB_WEBHOOK_SECRET": {
       "description": "Github Integration webhook secret"


app.json
```diff
@@ -33,7 +33,7 @@
       "description": "GitHub Integration ID"
     },
     "GITHUB_INTEGRATION_PEM": {
-      "description": "GitHub Integration private key (base64-encoded)"
+      "description": "GitHub Integration private key (download the file, encode it with base64, then paste the result here)"
```
Macarse: why don't you convert to base64 when you read the value? This sounds like a big pitfall for github users

notriddle: A PEM file has newlines in it, but a Heroku environment variable cannot have newlines in it.

Macarse: oh that sucks :(
Thanks for the explanation.

Macarse:
👍

roni-frantchi:
Thanks @notriddle.
Looks good to me.

notriddle:
bors r=roni-frantchi,Macarse

PR #152, opened 2017-04-13
Move the worker processes (Batcher, etc) into an app

notriddle:
There are a couple advantages to splitting up the code like this:

* Different OTP apps can have different distributed operation policies. For example, we can have a single instance of bors_worker with a fallback, while having two independent instances of bors_frontend.

* There are no circular dependencies allowed. An app lower in the dependency tree can't call a function defined higher in the tree.

* It controls which parts of bors depend on various external ones. Being sure where Phoenix stuff can be touched should help with migrating to Phoenix 1.3.

### DIFF
--- a/apps/bors_database/lib/project.ex
+++ b/apps/bors_database/lib/project.ex
@@ -9,6 +9,20 @@ defmodule BorsNG.Database.Project do
 
   @type t :: %Project{}
 
+  @pubsub_name Application.get_env(:bors_database, :pubsub)[:name]
+
+  @doc """
+  After modifying the underlying model,
+  call this to notify the UI.
+  """
+  def ping!(project_id) when not is_binary(project_id) do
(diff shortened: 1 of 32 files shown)


apps/bors_database/lib/project.ex
```diff
@@ -9,6 +9,20 @@ defmodule BorsNG.Database.Project do
...
   @type t :: %Project{}
 
+  @pubsub_name Application.get_env(:bors_database, :pubsub)[:name]
+
+  @doc """
+  After modifying the underlying model,
+  call this to notify the UI.
+  """
+  def ping!(project_id) when not is_binary(project_id) do
+    ping!("#{project_id}")
```
Macarse: replace with `ping!(to_string(project_id))`

Macarse:
👍

PR #170, opened 2017-04-24
Support BitBucket

Macarse:


apps/bors_frontend/web/controllers/webhook_controller.ex
```diff
@@ -79,11 +80,30 @@ defmodule BorsNG.WebhookController do
...
+        event = hd(get_req_header(conn, "x-github-event"))
+        do_webhook conn, "github", event
+
+      "bitbucket" ->
+        event = hd(get_req_header(conn, "x-event-key"))
+        Logger.info "Event: #{event}"
+        do_bitbucket_webhook conn, event
+
+        _ ->
+          conn |> send_resp(404, "")
```
Macarse: @notriddle not sure what to return here as response. Any thoughts?

apps/bors_frontend/web/controllers/webhook_controller.ex
```diff
@@ -79,11 +80,30 @@ defmodule BorsNG.WebhookController do
...
+    end
+
+    conn |> send_resp(200, "")
+  end
+
+  def do_bitbucket_webhook(conn, "pullrequest:comment") do
+    comment = conn.body_params["comment"]
+    Logger.info "comment: #{comment}"
+     if String.contains?(comment, "bors") do
+       Logger.info "Start command: #{comment}"
```
Macarse: @notriddle any reason why you don't do something like this before starting a command?

Macarse:
@notriddle can bors merge against branches which are not master?

apps/bors_frontend/web/controllers/webhook_controller.ex
notriddle: Probably 404, yes, since the provider is part of the URL.

apps/bors_frontend/web/controllers/webhook_controller.ex
```diff
@@ -79,11 +80,30 @@ defmodule BorsNG.WebhookController do
...
+          conn |> send_resp(404, "")
+    end
+
+    conn |> send_resp(200, "")
+  end
+
+  def do_bitbucket_webhook(conn, "pullrequest:comment") do
+    comment = conn.body_params["comment"]
+    Logger.info "comment: #{comment}"
+     if String.contains?(comment, "bors") do
```
Macarse: @notriddle any reason why you don't do something like this before starting a command?

notriddle:
> @notriddle can bors merge against branches which are not master?

It's not currently supported, but it's the current feature I'm working on.

apps/bors_frontend/web/controllers/webhook_controller.ex
notriddle: It didn't seem necessary, since every webhook that gets sent is already logged and available in the GitHub Webhook Configuration page.

notriddle:
This is going to be awhile. Probably better to close the PR now, then reopen it once there's more to review.

Issue #141, opened 2017-04-06
Add parallel building


notriddle:
Instead of batching stuff together, run every item in the backlog immediately with everything ahead of it merged in. This has a couple nice side-effects:

  * It reduces the time taken from O(E*log(N)) to O(E).
  * It results in a history that doesn't have arbitrary octopus merges.

On the other hand...

  * It increases the amount of builds run from O(E*log(N)) to O(E*N)
  * The time decrease only applies if you have an arbitrarily high number of build nodes. If you run out of nodes, it jumps to O(N) again.

So we do want to provide both.

Macarse:
Here's a nice diagram explanation of the algorithm: https://docs.openstack.org/infra/zuul/gating.html#testing-in-parallel

notriddle:
I should probably mention, just for completeness's sake, that this is going to be *awhile*.

Making good on parallel building requires near-bottomless build capacity and the ability for bors to insta-cancel builds. Travis limits free projects to something like five concurrent builds, which would be enough for bors-ng/bors-ng itself, but isn't going to cut it for rust-lang/rust.

Maybe we could use Travis's "auto-cancel" feature to drop known-bad builds from the queue? That could work, but I'd rather have something where bors can proactively throw errors if it's misconfigured. Being easy to set up matters.

PR #171, opened 2017-04-28
Allow each patch to have its own target branch

### LINKED ISSUE #136: Support pull requests against branches that aren't "master"
This requires bors to segregate batches. Maybe they should have completely separate Batcher processes?

notriddle:
Fixes #136

### DIFF
--- a/apps/bors_worker/test/attemptor_test.exs
+++ b/apps/bors_worker/test/attemptor_test.exs
@@ -22,7 +22,7 @@ defmodule BorsNG.Worker.AttemptorTest do
     {:ok, inst: inst, proj: proj}
   end
 
-  def new_patch(proj, pr_xref, commit \\ "N") do
+  def new_patch(proj, pr_xref, commit) do
     %Patch{
       project_id: proj.id,
       pr_xref: pr_xref,
@@ -67,7 +67,7 @@ defmodule BorsNG.Worker.AttemptorTest do
         statuses: %{"iniN" => []},
         files: %{"trying" => %{".travis.yml" => ""}}
       }})
-    patch = new_patch(proj, 1)
+    patch = new_patch(proj, 1, "N")
     Attemptor.handle_cast({:tried, patch.id, "test"}, proj.id)
     [status] = Repo.all(AttemptStatus)
     assert status.identifier == "continuous-integration/travis-ci/push"
@@ -81,7 +81,7 @@ defmodule BorsNG.Worker.AttemptorTest do
         statuses: %{"iniN" => []},
         files: %{"trying" => %{".travis.yml" => "", "appveyor.yml" => ""}}
       }})
-    patch = new_patch(proj, 1)
+    patch = new_patch(proj, 1, "N")
     Attemptor.handle_cast({:tried, patch.id, "test"}, proj.id)
     statuses = Repo.all(AttemptStatus)
     assert Enum.any?(statuses,
@@ -99,7 +99,7 @@ defmodule BorsNG.Worker.AttemptorTest do
         statuses: %{"iniN" => []},
         files: %{"trying" => %{"bors.toml" => ~s/status = [ "ci" ]/}}
       }})
-    patch = new_patch(proj, 1)
+    patch = new_patch(proj, 1, "N")
     Attemptor.handle_cast({:tried, patch.id, "test"}, proj.id)
     assert GitHub.ServerMock.get_state() == %{
       {{:installation, 91}, 14} => %{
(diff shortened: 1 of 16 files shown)


notriddle:
WIP: I need to write more test cases for this, first.

Macarse:
@notriddle two big questions:

1) I am not seeing in the PR is something to support the use case of a PR being updated and the target branch being changed.

Example:
I go to https://github.com/bors-ng/bors-ng/pull/170 and change the target branch from `bitbucket` to `master`.

Does that trigger any hook on github?
What should be the outcome of changing the target branch after doing `bors r+` ?

I also noticed that Github calls that `base branch` instead of target branch. Should we edit?

2) What happens if you do `bors r+` on two PRs at the same time, one going to `master` and the other going to `x`?
There is still a notion of a single `integration` branch instead of something like `integration_master`, `integration_x`. Also, there is a unique instance of a batcher, right?

apps/bors_worker/test/attemptor_test.exs
```diff
@@ -16,13 +16,26 @@ defmodule BorsNG.Worker.AttemptorTest do
...
       installation_id: inst.id,
       repo_xref: 14,
-      master_branch: "master",
       staging_branch: "staging",
       trying_branch: "trying"}
     |> Repo.insert!()
     {:ok, inst: inst, proj: proj}
   end
 
+  def new_patch(proj, pr_xref, commit \\ "N") do
```
Macarse: why the default for commit is `N`?

notriddle:
> Does that trigger any hook on github?

It triggers a [`pull_request` event](https://developer.github.com/v3/activity/events/types/#pullrequestevent), with the "action" set to "edited". I need to update the webhook controller to use it.

> What happens if you do bors r+ on two PRs at the same time, one going to master and the other going to x?

Pull requests with different base branches will never be batched together. However, it does not run them concurrently; they'll just queue up. As you've already noticed, fixing that would require multiple integration branches.

apps/bors_worker/test/attemptor_test.exs
notriddle: Me being lazy.

apps/bors_worker/test/batcher/batcher_test.exs
```diff
@@ -438,6 +493,92 @@ defmodule BorsNG.Worker.BatcherTest do
...
+      {{:installation, 91}, 14} => %{
+        branches: %{"master" => "ini", "staging" => "", "staging.tmp" => ""},
+        comments: %{1 => [], 2 => []},
+        statuses: %{},
+        files: %{"staging" => %{"bors.toml" => ~s/status = [ "ci" ]/}}
+      }})
+    patch = %Patch{
+      project_id: proj.id,
+      pr_xref: 1,
+      commit: "N",
```
Macarse: I read your test case like 3 times thinking so if release has 0 commits and master has N... 
(facepalm)

Macarse:
> Pull requests with different base branches will never be batched together. However, it does not run them concurrently; they'll just queue up. As you've already noticed, fixing that would require multiple integration branches.

Isn't there a reason why you wouldn't do it?

notriddle:
> Is there a reason why you wouldn't do it?

  * Ease of setup is the biggest reason. Right now, project maintainers only need to whitelist "staging" and "trying." If I added a separate staging branch for each target branch, then they'd need to whitelist "staging.master," "staging.beta," "staging.stable," "trying.master," "trying.beta," and "trying.stable."

  * The performance improvement seems pretty marginal, assuming a Rust-like setup with a "master" branch that gets the bulk of the traffic, plus feature-frozen "beta" and "stable" branches.

  * If you're planning something more elaborate, then it's probably a good idea to use multiple repositories. [Look at all the "team forks" of linux on git.kernel.org](https://git.kernel.org/), for example.

notriddle:
> Also, there is a unique instance of a batcher, right?

Each project (really GitHub repository) has its own, unique batcher. Yes.

Macarse:
> The performance improvement seems pretty marginal, assuming a Rust-like setup with a "master" branch that gets the bulk of the traffic, plus feature-frozen "beta" and "stable" branches.

My biggest concern is blocking the master traffic while building a one off. Also feature branches might want to setup their own queue. 

Examples of how my repo works:

 - Master => bulk of the traffic as you mentioned
 - Release => few changes and I would prefer them to be parallel to master
 - FeatureX => long live feature branch which would want a separate queue. Sometimes this features have a different builder and add steps which are not needed by master. They can't live in a separate repo since it needs to rebase again master periodically.

Setup will depend on projects and I think bors should be flexible.

I guess we can discuss more about this when developing https://github.com/bors-ng/bors-ng/issues/141 That feature would require parallel building.

👍

notriddle:
bors r=Macarse

PR #210, opened 2017-06-09
Add the `cut-body-after` configuration option

### LINKED ISSUE #200: Crop off point for pull requests
Add a bors.toml directive to the effect of:

```toml
cut-description-after = "<!-- Reviewable:start -->"
```

If present, then when bors writes PR bodies into its merge commit, it will ignore the text blob and everything after it.

notriddle:
Fixes #200

### DIFF
--- a/apps/bors_github/lib/github.ex
+++ b/apps/bors_github/lib/github.ex
@@ -62,7 +62,7 @@ defmodule BorsNG.GitHub do
     from: bitstring,
     to: bitstring,
     commit_message: bitstring,
-    }) :: %{commit: bitstring, tree: binary} | :conflict
+    }) :: %{commit: binary, tree: binary} | :conflict
   def merge_branch!(repo_conn, info) do
     {:ok, commit} = GenServer.call(
       BorsNG.GitHub,
--- a/apps/bors_worker/lib/batcher/message.ex
+++ b/apps/bors_worker/lib/batcher/message.ex
@@ -71,18 +71,27 @@ defmodule BorsNG.Worker.Batcher.Message do
     "#{acc}\n  * #{status_link}"
   end
 
-  def generate_commit_message(patch_links) do
+  def generate_commit_message(patch_links, cut_body_after) do
     patch_links = Enum.sort_by(patch_links, &(&1.patch.pr_xref))
     commit_title = Enum.reduce(patch_links,
       "Merge", &"#{&2} \##{&1.patch.pr_xref}")
     commit_body = Enum.reduce(patch_links, "", fn link, acc ->
+      body = cut_body(link.patch.body, cut_body_after)
       """
       #{acc}
       #{link.patch.pr_xref}: #{link.patch.title} r=#{link.reviewer}
-      #{link.patch.body}
+
+      #{body}
       """
     end)
     "#{commit_title}\n#{commit_body}"
+  end
+
+  def cut_body(body, nil), do: body
+  def cut_body(body, cut) do
+    body
+    |> String.splitter(cut)
+    |> Enum.at(0)
   end
 
   def generate_bors_toml_error(:parse_failed) do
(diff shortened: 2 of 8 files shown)


notriddle:
This should be in a reviewable state now.

apps/bors_github/lib/github.ex
```diff
@@ -62,7 +62,7 @@ defmodule BorsNG.GitHub do
     from: bitstring,
     to: bitstring,
     commit_message: bitstring,
-    }) :: %{commit: bitstring, tree: binary} | :conflict
+    }) :: %{commit: binary, tree: binary} | :conflict
```
Macarse: Why was this change needed?

apps/bors_worker/test/batcher/message_test.exs
```diff
@@ -50,16 +50,51 @@ defmodule BorsNG.Worker.BatcherMessageTest do
...
+    - [ ] Run the linter
+    - [ ] Run any new or changed tests
+    - [ ] This PR fixes #___ (fill in if it exists)
+    - [ ] Make sure your commit messages make sense
+    """
+    patches = [
+      %{patch: %{pr_xref: 1, title: title, body: body}, reviewer: "bill"} ]
+    actual_message = Message.generate_commit_message(
+      patches,
+      "\n\n<!-- boilerplate follows -->")
```
Macarse: Would bors allow to use `\n` inside the cut off token?

apps/bors_worker/test/batcher/message_test.exs
```diff
@@ -50,16 +50,51 @@ defmodule BorsNG.Worker.BatcherMessageTest do
...
+    assert expected_message == actual_message
+  end
 end
```
Macarse: For completeness, I would add a test for more than one match and another with no matches at all.

apps/bors_github/lib/github.ex
notriddle: It actually wasn't needed. The reason I did it is because [`binary` is a more specific typespec than `bitstring`](https://hexdocs.pm/elixir/typespecs.html#types-and-their-syntax). Namely, if you assert that the type is `binary`, you assert that the number if bits in the bitstring is a multiple of 8 (which is always true for UTF-8).

apps/bors_worker/test/batcher/message_test.exs
notriddle: Yeah, it would.

apps/bors_worker/test/batcher/message_test.exs
notriddle: Done.

notriddle:
BTW, you probably don't want to merge it until I've squashed the `fixup!` commit into the main commit.

apps/bors_worker/lib/batcher/message.ex
```diff
@@ -71,20 +71,30 @@ defmodule BorsNG.Worker.Batcher.Message do
...
+      #{body}
       """
     end)
     "#{commit_title}\n#{commit_body}"
   end
 
+  def cut_body(body, nil), do: body
+  def cut_body(body, cut) do
+    body
+    |> String.splitter(cut)
```
Macarse: while searching about `String.splitter` I found this: https://github.com/elixir-lang/elixir/issues/6148

notriddle:
bors r= @Macarse

notriddle:
As discussed in IRC, switched to `Enum.at(0)` instead of a combination of `Enum.take(1) |> hd()`

apps/bors_worker/lib/batcher/message.ex
```diff
@@ -71,20 +71,30 @@ defmodule BorsNG.Worker.Batcher.Message do
...
+    |> Enum.take(1)
+    |> hd()
```
Macarse: Talked to @notriddle privately. this can be done without the `hd()`. Using `Enum.at(0)`.

PR #211, opened 2017-06-10
Improve "permission denied" handling

notriddle:
* When a non-reviewer attempts to post a review, put a link to the dashboard to add them as a reviewer
* When a non-reviewer tries to open the settings page (or click the aforementioned link), respond with a 403 instead of a 500


notriddle:
@Macarse You're quick!

PR #214, opened 2017-06-14
Make the command parser more line-oriented

### LINKED ISSUE #173: Do not act on "commands" that are in code quotes, code blocks, or block quotes
The way bors responded to [the meta-discussion in this PR](https://github.com/bors-ng/bors-ng/pull/171) was just comical.

notriddle:
This is a [breaking change], though it's designed to match common usage. It also fixes #173, and it adds autocorrections for common typing errors (with warnings).

### DIFF
--- a/apps/bors_frontend/config/config.exs
+++ b/apps/bors_frontend/config/config.exs
@@ -9,11 +9,8 @@ config :bors_frontend, ecto_repos: []
 
 # General application configuration
 config :bors_frontend, BorsNG,
-  activation_phrase: "bors r+",
-  activation_by_phrase: "bors r=",
-  deactivation_phrase: "bors r-",
+  command_trigger: "bors ",
   home_url: "https://bors-ng.github.io/",
-  try_phrase: "bors try",
   github_integration_url:
     "https://github.com/integrations/bors/installations/new",
   allow_private_repos: System.get_env("ALLOW_PRIVATE_REPOS") == "true"
--- a/apps/bors_frontend/web/command.ex
+++ b/apps/bors_frontend/web/command.ex
@@ -38,14 +38,8 @@ defmodule BorsNG.Command do
     patch: nil,
     comment: "")
 
-  @activation_phrase(
-    Application.get_env(:bors_frontend, BorsNG)[:activation_phrase])
-  @activation_by_phrase(
-    Application.get_env(:bors_frontend, BorsNG)[:activation_by_phrase])
-  @deactivation_phrase(
-    Application.get_env(:bors_frontend, BorsNG)[:deactivation_phrase])
-  @try_phrase(
-    Application.get_env(:bors_frontend, BorsNG)[:try_phrase])
+  @command_trigger(
+    Application.get_env(:bors_frontend, BorsNG)[:command_trigger])
 
   @type t :: %BorsNG.Command{
     project: Project.t,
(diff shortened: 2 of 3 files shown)


apps/bors_frontend/config/config.exs
```diff
@@ -9,11 +9,11 @@ config :bors_frontend, ecto_repos: []
 
 # General application configuration
 config :bors_frontend, BorsNG,
-  activation_phrase: "bors r+",
-  activation_by_phrase: "bors r=",
-  deactivation_phrase: "bors r-",
+  activation_phrases: ["bors r+", "bors +r"],
+  activation_by_phrases: ["bors r="],
```
Macarse: why there is not `bors =r`?

apps/bors_frontend/web/command.ex
```diff
@@ -1,3 +1,18 @@
+defmodule BorsNG.Command.MultiParse do
+  defmacro multi_parse(phrases, param, [do: result]) do
```
Macarse: why a macro and not a function?

apps/bors_frontend/web/command.ex
```diff
@@ -1,3 +1,18 @@
...
+    defs = :bors_frontend
+    |> Application.get_env(BorsNG)
+    |> Keyword.get(phrases)
+    |> Enum.map(fn phrase ->
+      quote do
+        def parse(unquote(phrase) <> unquote(param)) do
```
Macarse: I do not understand this, I guess it's macro stuff

Macarse:
Honestly, I do not like this change. 
Why instead of supporting wrong typed commands we show a comment in the PR saying something like:

`@user: Did you mean bors r+?`

apps/bors_frontend/config/config.exs
notriddle: The `=` sign connects the nickname with the `r`. The `activation_by_phrases` allow me to type in something like "r=Macarse" (I'm typing the command, but you get recorded as the reviewer). "=rMacarse" is dumb.

apps/bors_frontend/web/command.ex
notriddle: It's a macro because it's generating other functions. Thinking further about it, that's probably a dumb way to go about this.

apps/bors_frontend/web/command.ex
```diff
@@ -223,4 +229,10 @@ defmodule BorsNG.Command do
     attemptor = Attemptor.Registry.get(c.project.id)
     Attemptor.tried(attemptor, c.patch.id, arguments)
   end
+  def run(c, {:autocorrect, command}, _) do
+    c.project.repo_xref
+    |> Project.installation_connection(Repo)
+    |> GitHub.post_comment!(
+      c.pr_xref, ~s/Auto-corrected to "#{command}"/)
```
Macarse: With this change there will be a msg in the pr saying:
"It's running because it was autocorrected"

I was leaning towards not executing anything at all when there is a typo and letting the user know. The same way git works:
```
$ git statys
git: 'statys' is not a git command. See 'git --help'.

Did you mean this?
	status
```

Making the typo trigger an action is just incrementing the commands inside bors. If you add this, someone might come in the future and would also like to have: bor+, ++, etc.

notriddle:
Yeah, I see what you mean.

notriddle:
Now that the change has been made, does it otherwise look good to you?

notriddle:
bors r=Macarse

PR #218, opened 2017-06-21
Allow reviewers to give permission to review a single PR

### LINKED ISSUE #207: Allow delegating approval
Sometimes I'd like to delegate the `bors r+` to the PR author (for example, I request a squash before merging).

It could look like this:

```
bors delegate=AnotherUser
```

notriddle:
Fixes #207

### DIFF
--- a/apps/bors_database/lib/context.ex
+++ b/apps/bors_database/lib/context.ex
@@ -1,4 +1,4 @@
-defmodule BorsNG.Database.Model do
+defmodule BorsNG.Database.Context do
   @moduledoc """
   A module that keeps using definitions for models.
 
@@ -14,8 +14,6 @@ defmodule BorsNG.Database.Model do
   """
   defmacro __using__(_) do
     quote do
-      use Ecto.Schema
-
       import Ecto
       import Ecto.Changeset
       import Ecto.Query
@@ -29,6 +27,7 @@ defmodule BorsNG.Database.Model do
       alias BorsNG.Database.LinkUserProject
       alias BorsNG.Database.Patch
       alias BorsNG.Database.Project
+      alias BorsNG.Database.Repo
       alias BorsNG.Database.Status
       alias BorsNG.Database.User
       alias BorsNG.Database.UserPatchDelegation
--- a/apps/bors_database/lib/context/permission.ex
+++ b/apps/bors_database/lib/context/permission.ex
@@ -8,16 +8,28 @@ defmodule BorsNG.Database.Context.Permission do
 
   use BorsNG.Database.Context
 
-  def user_has_permission_to_approve_patch(user, patch) do
+  def permission_to_approve_patch?(user, patch) do
     %User{id: user_id} = user
     %Patch{id: patch_id, project_id: project_id} = patch
-    delegated = UserPatchDelegation
-    |> where([d], d.user_id == ^user_id and d.patch_id == ^patch_id)
-    |> Repo.one()
-    linked = LinkUserProject
+    project_reviewer?(user_id, project_id) or
+      patch_delegated_reviewer?(user_id, patch_id)
+  end
+
+  def project_reviewer?(user_id, project_id) do
+    LinkUserProject
     |> where([l], l.user_id == ^user_id and l.project_id == ^project_id)
     |> Repo.one()
-    (not is_nil linked) or (not is_nil delegated)
+    |> is_nil()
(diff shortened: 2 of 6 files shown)


torkleyy:
@notriddle Thank you for implementing this feature :+1: 

I'm sorry I'm not suitable for reviewing this PR, because I'm neither familiar with this code base nor the Euphoria Programming Language.

Macarse:
@notriddle I will review right now.

apps/bors_database/lib/context.ex
```diff
@@ -0,0 +1,36 @@
...
+
+      use BorsNG.Database.Model
+  """
+
+  # Do NOT define functions inside the quoted expression.
+
+  @doc """
+  When used, add common imports for models.
+  """
+  defmacro __using__(_) do
```
Macarse: macro for imports?

apps/bors_database/lib/context/permission.ex
```diff
@@ -0,0 +1,28 @@
...
+  """
+
+  use BorsNG.Database.Context
+
+  def user_has_permission_to_approve_patch(user, patch) do
+    %User{id: user_id} = user
+    %Patch{id: patch_id, project_id: project_id} = patch
+    delegated = UserPatchDelegation
+    |> where([d], d.user_id == ^user_id and d.patch_id == ^patch_id)
+    |> Repo.one()
```
Macarse: If `delegated` is not nil you don't need to do the second query, right?

apps/bors_database/lib/context/permission.ex
```diff
@@ -0,0 +1,28 @@
...
+    linked = LinkUserProject
+    |> where([l], l.user_id == ^user_id and l.project_id == ^project_id)
+    |> Repo.one()
+    (not is_nil linked) or (not is_nil delegated)
```
Macarse: I would like to see some spacing between lines. I thought it was a single query. Perhaps something like:

```
%User{id: user_id} = user
    %Patch{id: patch_id, project_id: project_id} = patch

    delegated = UserPatchDelegation
    |> where([d], d.user_id == ^user_id and d.patch_id == ^patch_id)
    |> Repo.one()

    linked = LinkUserProject
    |> where([l], l.user_id == ^user_id and l.project_id == ^project_id)
    |> Repo.one()

    (not is_nil linked) or (not is_nil delegated)
```

mmun:
It makes sense to me. I didn't quite get the motivating example in #207 though. Are they assuming that the person will squash correctly? Or in worse cases, is there anything stopping someone from adding a nefarious commit?

apps/bors_database/lib/context/permission.ex
Macarse: if you do this change, I guess it makes sense to do it in the opposite order `LinkUserProject` first, `UserPatchDelegation` second.

notriddle:
The motivating example in that issue is pretty vague and short, especially since there's really two cases I would use this:

1. For someone who's being "trained" for the code review job, you'll want to delegate them a few PRs before giving them the review bit for real. Personally, I think this is the most important reason to have delegation.
2. For someone who's already a regular contributor, and you know that they know how to correctly execute on what you're asking for. The squash-and-merge example is just a particular case of this.

In neither of these will you be giving the delegate bit to a new contributor. You'd also probably want to keep an eye on what they do with that permission so that you know whether to trust them with it again.

apps/bors_database/lib/context.ex
notriddle: Yeah, that's exactly what it is.

notriddle:
@Macarse That should fix up all of your complaints.

Macarse:
@notriddle not complaints, recommendations ;)

apps/bors_database/lib/context/permission.ex
```diff
@@ -0,0 +1,40 @@
...
+    project_reviewer?(user_id, project_id) or
+      patch_delegated_reviewer?(user_id, patch_id)
+  end
+
+  defp project_reviewer?(user_id, project_id) do
+    LinkUserProject
+    |> where([l], l.user_id == ^user_id and l.project_id == ^project_id)
+    |> Repo.one()
+    |> is_nil()
+    # elixirc squawks about unary operators if the module is left off.
```
Macarse: interesting

notriddle:
Deployed.

PR #223, opened 2017-07-11
Ignore non-create comment commands

### LINKED ISSUE #196: Wrong "not awaiting review" comment
For the PR i've r+'ed, bors replied that it does not await a review, although it was the only r+ on this PR:

https://github.com/intellij-rust/intellij-rust/pull/1282#issuecomment-304500234

notriddle:
Fixes #196


Macarse:
@notriddle so based on this fix, if I edit a comment adding a command it will not be executed, right?

notriddle:
Right.

PR #224, opened 2017-07-11
fix(webhooks): rename integration events

### LINKED ISSUE #189: Switch the webhook controller from Integration events to App events
According to GitHub's recent [blog post](https://developer.github.com/changes/2017-05-22-github-apps-production-ship/):

> With the name change to GitHub Apps, there are some changes to webhook events, payloads and API responses.
>
> The `integration_installation` and `integration_installation_repositories` event names, used in the Early Access period, have been deprecated in favor of `installation` and `installation_repositories`. All GitHub Apps receive these events to their webhook by default, so all applications are affected by this change.

The code to handle these events is in the [`BorsNG.WebhookController`](https://github.com/bors-ng/bors-ng/blob/72e27c02dffa0698785c97645e3999b0772c526f/apps/bors_frontend/web/controllers/webhook_controller.ex#L93) module. The events are not semantically different; you just need to change the name. It helps that we never used the `integration_id` anyway; we only use the `installation_id`, which has not been renamed.

You'll also need to fix the doctest at the top of the file.

jniles:
GitHub's API migrated from integration events to app events, a semantic
change.  The webhooks controller is updated to reflect the naming
convention now used by GitHub.

See the following blog for more infor:
https://developer.github.com/changes/2017-05-22-github-apps-production-ship/

Closes #189.

### DIFF
--- a/apps/bors_frontend/web/controllers/webhook_controller.ex
+++ b/apps/bors_frontend/web/controllers/webhook_controller.ex
@@ -46,7 +46,7 @@ defmodule BorsNG.WebhookController do
       ...>       "id" => 6,
       ...>       "login" => "bors-fanboi",
       ...>       "avatar_url" => "" },
-      ...>     "action" => "created" }}, "github", "integration_installation")
+      ...>     "action" => "created" }}, "github", "installation")
       iex> proj = Database.Repo.get_by!(Database.Project, repo_xref: 14)
       iex> proj.name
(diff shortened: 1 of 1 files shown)


notriddle:
Thanks for working on this! I know how time-consuming it can be to get familiar with a new codebase, especially one that's so heavily async as this one.

bors r+

PR #226, opened 2017-07-13
Fix up some look-and-feel papercuts

### LINKED ISSUE #159: Visited links look surprisingly on the dashboard
Hi! Currently, links in the dashboard use custom colors for visited/not visited states:

links

This is a bit confusing because the difference in color is really pronounced and looks like an indication of the failure/success status because of the reddish tint of the visited link. Perhaps visited/not visited colors should not differ that much in general? Alternatively, it may be worthwhile to remove the "visited" marker from these particular links altogether, it does not seem to add much value, but does add some noise.

notriddle:
A lot of this is duplicate of bors-ng/bors-ng.github.io#14.

Also fixes #159

### DIFF
--- a/apps/bors_frontend/web/static/css/app.css
+++ b/apps/bors_frontend/web/static/css/app.css
@@ -79,7 +79,7 @@ footer {
   color: #33b5e5;
 }
 :visited {
-  color: #ae4d76;
+  color: #ae3d96;
 }
 :visited:focus {
   color: #b533e5;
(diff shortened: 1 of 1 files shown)


apps/bors_frontend/web/static/css/app.css
```diff
@@ -218,8 +232,14 @@ footer {
...
-.toolbar-button--remove {
-  background-color: #df3600;
+.toolbar-item--add:focus {
+  background: #0A0;
+}
+.toolbar-item--remove {
+  background-color: #cf3600;
+}
+.toolbar-item--remove:focus {
+  background: #F22;
```
Macarse: reviewing css is harder than reviewing elixir...

Issue #227, opened 2017-07-14
Create pages for looking at individual batches


notriddle:
There should be a page where users can look at the status for a single, ongoing batch.

* It should show its current status, of course.
* It should show how long it's been running for.
* It should show the status of each of its sub-steps (the contexts, like Travis, AppVeyor, CircleCI, BuildBot).
* It should be linked to by bors' own context (the one at the bottom of the PR screen) as well as the History tab (PR #history).

PR #228, opened 2017-07-14
Give repository owners more visibility into their history

### LINKED ISSUE #216: Include repository source of logs in logs
https://elixirforum.com/t/logger-categorized-by-supervision-subtrees/6127/1

notriddle:
This adds a "History" tab, where crash reports and past batches can be seen

This will log crash reports (so this PR fixes #216, though none of that data exists yet) and allows all the information about past batches (which we have been keeping, BTW, so this tab should already be populated for most long-standing repos).

Next step: tackle #227

### DIFF
--- a/apps/bors_database/lib/crash.ex
+++ b/apps/bors_database/lib/crash.ex
@@ -1,0 +1,34 @@
+defmodule BorsNG.Database.Crash do
+  @moduledoc """
+  Records of the 
+  Corresponds to a pull request in GitHub.
+
+  A closed patch may not be r+'ed,
+  nor can a patch associated with a completed batch be r+'ed again,
+  though a patch may be merged and r+'ed at the same time.
+  """
+
+  use BorsNG.Database.Model
+
+  @type t :: %Crash{}
+
+  schema "crashes" do
+    belongs_to :project, Project
+    field :crash, :string
+    field :component, :string
+    timestamps()
+  end
+
+  @doc """
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:crash, :component])
+  end
+
+  def all_for_project(project_id) do
+    from c in Crash,
+      where: c.project_id == ^project_id
+  end
+end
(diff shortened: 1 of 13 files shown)


apps/bors_database/lib/crash.ex
```diff
@@ -0,0 +1,34 @@
...
+  Builds a changeset based on the `struct` and `params`.
+  """
+  def changeset(struct, params \\ %{}) do
+    struct
+    |> cast(params, [:crash, :component])
+  end
+
+  def all_for_project(project_id) do
+    from c in Crash,
+      where: c.project_id == ^project_id
```
Macarse: is the `^` actually needed?

apps/bors_frontend/web/controllers/project_controller.ex
```diff
@@ -89,6 +90,25 @@ defmodule BorsNG.ProjectController do
       update_branches: Project.changeset_branches(project)
   end
 
+  def log(_, :ro, _, _), do: raise BorsNG.PermissionDeniedError
```
Macarse: `:ro` is not allowed and `:rw` is? Sounds it should be the other way around

apps/bors_database/lib/crash.ex
notriddle: Yeah, it's part of the Ecto syntax. If you don't put it there, it'll complain about `project_id` not being declared within the query (and suggest using the `^` syntax if it's a local variable that you want to interpolate).

apps/bors_frontend/web/controllers/project_controller.ex
notriddle: `:rw` is a superset of `:ro`. I picked the more stringent permission mostly to be conservative.

apps/bors_database/lib/crash.ex
notriddle: https://hexdocs.pm/ecto/Ecto.Query.html gives a description of the syntax for the `from` macro.

notriddle:
Since you said it was approved, I'll go with that.

bors r= @Macarse

PR #231, opened 2017-07-17
Fix crash when trying to bisect a batch

notriddle:
Fixes a crash that happened to badboy, and probably others.

Regression introduced by #121

### DIFF
--- a/apps/bors_worker/lib/batcher.ex
+++ b/apps/bors_worker/lib/batcher.ex
@@ -273,7 +273,7 @@ defmodule BorsNG.Worker.Batcher do
   defp start_waiting_merged_batch(batch, patch_links, _base, :conflict) do
     repo_conn = get_repo_conn(batch.project)
     patches = Enum.map(patch_links, &(&1.patch))
-    state = bisect(patches, batch)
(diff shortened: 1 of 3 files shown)


apps/bors_worker/lib/batcher.ex
```diff
@@ -344,19 +344,21 @@ defmodule BorsNG.Worker.Batcher do
     project = batch.project
     repo_conn = get_repo_conn(project)
     erred = Enum.filter(statuses, &(&1.state == Status.numberize_state(:error)))
-    patches = batch.id
-    |> Patch.all_for_batch()
+    patch_links = batch.id
+    |> LinkPatchBatch.from_batch()
     |> Repo.all()
-    state = bisect(patches, batch)
+    patches = Enum.map(patch_links, &(&1.patch))
```
Macarse: `&(&1.patch)`? what's that?

notriddle: Shorthand for `fn link -> link.patch end`

It's an anonymous function.

notriddle:
Dogma was complaining.

bors r=Macarse

PR #232, opened 2017-07-19
accept bors commands with a colon

### LINKED ISSUE #219: Accept bors commands with a colon
https://github.com/georust/rust-geo/pull/120

khodzha:
pr for issue #219
not sure if _parse is appropriate name for a function tho :thinking: 

*Edit by @notriddle: fixes #219*

### DIFF
--- a/apps/bors_frontend/config/config.exs
+++ b/apps/bors_frontend/config/config.exs
@@ -9,7 +9,7 @@ config :bors_frontend, ecto_repos: []
 
 # General application configuration
 config :bors_frontend, BorsNG,
-  command_trigger: "bors ",
+  command_trigger: "bors",
   home_url: "https://bors-ng.github.io/",
   github_integration_url:
     "https://github.com/integrations/bors/installations/new",
--- a/apps/bors_frontend/test/command_test.exs
+++ b/apps/bors_frontend/test/command_test.exs
@@ -24,6 +24,10 @@ defmodule BorsNG.CommandTest do
     assert [:deactivate] == Command.parse("bors r-")
   end
 
+  test "accept command with colon after it" do
+    assert [{:try, ""}] == Command.parse("bors: try")
+  end
+
   test "accept the try command with an argument" do
     assert [{:try, "-layout"}] == Command.parse("bors try-layout")
   end
--- a/apps/bors_frontend/web/command.ex
+++ b/apps/bors_frontend/web/command.ex
@@ -107,28 +107,28 @@ defmodule BorsNG.Command do
     |> String.splitter("\n")
     |> Enum.flat_map(fn
       @command_trigger <> " " <> cmd ->
-        _parse(cmd)
+        parse_cmd(cmd)
       @command_trigger <> ": " <> cmd ->
-        _parse(cmd)
+        parse_cmd(cmd)
       _ -> []
     end)
   end
 
-  def _parse(cmd) do
+  def parse_cmd(cmd) do
     cmd
     |> String.trim()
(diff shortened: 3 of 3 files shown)


Macarse:
We are already reporting back when the command was nor parsed correctly. I think this should be the same. Instead of accepting `bors: +r` bors should post saying: "Did you mean ..."

see https://github.com/bors-ng/bors-ng/pull/214

notriddle:
If you want suggestions, you could name the new function `parse_cmd` and rename the current function with that name to `parse_trimmed_cmd`.

notriddle:
@Macarse

Unlike the case of actual misspelled commands, I'm not sure what the advantage of rejecting `bors: r+` is. It's obvious what the user means when they type that.

This PR only adds the `:` as a permitted delimiter between "bors" and the command. It doesn't make it accept `+r` as the review command.

Macarse:
@notriddle because you are extending the API. If `bors: r+` is accepted, I also want `bors; r+`.

Why do we want to add this? To make sure developers do not waste time waiting for a build that never started?
Considering this a misspelled command would notify the developer immediately.

khodzha:
im not actually a bors-ng user, yet `bors: r+` seems meaningful, `bors; r+` does not
but i agree that bloating allowed separators can be a problem

notriddle:
> Why do we want to add this? To make sure developers do not waste time waiting for a build that never started?

Mostly because it's a common thing I see homu users doing. I've already had to teach people who're migrating not to start the bot's name with an `@` sign and that the command has to go on its own line. The "own line" thing is a restriction that I'm still not sure if I'm comfortable with, but at least there I can name an actual tradeoff in its favor (it reduces false positives like when people talk *about* bors's commands).

Carrying the colon thing into perpetuity seems like such a tiny burden compared to all the other compatibility stuff that bors-ng does.

notriddle:
@khodzha And thanks for putting up with our last-minute bickering. Honestly, it's kind of embarrassing that we haven't already resolved this 😞

notriddle:
To summarize: I filed issue #219 because it seemed like a good idea to make bors-ng's syntax more like homu's. To reduce friction for people migrating to it.

notriddle:
@khodzha ... And thank you for putting up with the 80char line limit. 😁

Macarse:
> because it seemed like a good idea to make bors-ng's syntax more like homu's. To reduce friction for people migrating to it.

Posting a msg in the PR with the "Did you mean bors r+" also reduces the friction and teaches the user the correct way and unique way of executing actions.

@notriddle As mentioned in the previous PR, this is your call. You should be the one making the final decision on the API we support.

notriddle:
Sure, okay.

bors r+

Next time, though, I'm opening a B-RFC issue before putting API-affecting changes on E-easy. Dropping an API design discussion on a first-time contributor is not the kind of environment that I think anyone wants to create for first-timers.

notriddle:
And thanks for showing up and (essentially) beta-testing the decision-making process, @khodzha.

khodzha:
you are welcome! :smiley:

notriddle:
Welp, that's done. It should be getting deployed to the test instance in a few minutes, and then it'll be ready for everyone to use.

notriddle:
[And... there it goes!](https://github.com/notriddle/test_repo/pull/100)

Macarse:
@notriddle 100% agree.
@khodzha sorry for the discussion in your first PR! Second will go faster for sure 👍

khodzha:
no need for excuses, it was important thing to clear up

PR #233, opened 2017-07-20
Fix typo

Fuco1:


notriddle:
And that's how the process is supposed to work! :+1:

PR #234, opened 2017-07-20
bors ping implementation

### LINKED ISSUE #120: Add a "bors ping" command
See also: servo/homu#83

khodzha:
fixes #120 

not sure if i should add tests and where to do that

### DIFF
--- a/apps/bors_frontend/test/command_test.exs
+++ b/apps/bors_frontend/test/command_test.exs
@@ -1,9 +1,26 @@
 defmodule BorsNG.CommandTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case
 
   alias BorsNG.Command
+  alias BorsNG.Database.Installation
+  alias BorsNG.Database.Project
+  alias BorsNG.Database.Repo
+  alias BorsNG.GitHub
 
   doctest BorsNG.Command
+
+  setup do
+    :ok = Ecto.Adapters.SQL.Sandbox.checkout(Repo)
+
+    inst = %Installation{installation_xref: 91}
+    |> Repo.insert!()
+    proj = %Project{
+      installation_id: inst.id,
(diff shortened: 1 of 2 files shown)


notriddle:
If you want to add a test, it would probably go in [command_test.exs](https://github.com/bors-ng/bors-ng/blob/97720d8ee66bbdf182e00083423e7155333f5dcc/apps/bors_frontend/test/command_test.exs)

khodzha:
is testing only command parsing ok?
shouldnt i test run(c, :ping) behaviour?

notriddle:
Yeah, it's probably a bad thing that there aren't any test cases for that part of the command module. All of the stuff that's needed to make it work does exist, and you can see how to use it in the `batcher_test.exs` file.

You're going to need to remove the `, async: true` part of the `command_test.exs` file, so that it matches the batcher test. This is because it's interacting with the Global Mock GitHub Server.

Secondly, in your "run functionality test", you'll need to initialize the github mock with some example state (most of the batcher tests do it, so you can see an example), then call the run function, then assert on the new state that should result.

khodzha:
hmm, it looks like i would need a db connection to make this code:
```
    c.project.repo_xref
    |> Project.installation_connection(Repo)
```
work in test env. is it ok? seems like pulling a big dependency

notriddle:
Yes, you will need a database connection. Yes, the test env has a database connection; it needs it to be able to run a bunch of other tests anyway.

notriddle:
bors r+

[LOOKS GOOD TO ME](http://www.lgtm.in/i/ds92)

khodzha:
is
> not sure if i should add tests and where to do that
going to get into commit msg?

notriddle:
Yeah, it got into the commit message. Took me too long to parse your question; sorry about that.

PR #236, opened 2017-07-21
accept deprecated events

khodzha:
let deprecated `integration_installation` and `integration_installation_repositories` come and serve http 200 response instead of http 500

code for this can be deleted after November 22, 2017


notriddle:
bors r+

That works.

Issue #177, opened 2017-05-05
Option to delete branch automatically once it has been merged


matklad:
Feature request. 

Using bors-ng workflow results in tons of merged branches. It would be convenient if bors automatically deleted remote branch after it had been merged.

Macarse:
@notriddle I noticed you do this with your prs. You do it yourself through the cmd line, right?

notriddle:
I use the CLI to open Pull Requests, and you can delete a branch using the CLI with `git push origin :BRANCH_NAME`, but that's not the point.

The point is that you don't need a branch after it's merged, so bors should just delete it.

Macarse:
@notriddle oh, yes, it totally makes sense to me but  I wanted to understand how it was happening now.

PS: you can also use `git push --delete origin branch_name` which is the "new" way :)

notriddle:
Fixed by #237

PR #237, opened 2017-07-21
same repo branch deletion after succesfull merge

khodzha:


khodzha:
how should i test async behaviour?

khodzha:
this is for issue #177

notriddle:
You can poll the server until you get the result you want. Since it's all in-process and governed by the EVM's fair scheduler, it won't slow things down to poll without throttling.

khodzha:
and how do i fix dialyzer complaining about
> The pattern 'true' can never match the type 'false'

since @delete_merged_branches set in config file to false?

notriddle:
The config.exs files are global for the entire instance of bors-ng.

I think it would be a lot better if you pulled that setting out of the bors.toml. That way, different users of the public instance can pick their setting.

notriddle:
That'll also fix the dialyzer warning.

apps/bors_worker/lib/batcher.ex
```diff
@@ -337,6 +342,7 @@ defmodule BorsNG.Worker.Batcher do
     patches = batch.id
     |> Patch.all_for_batch()
     |> Repo.all()
+    Process.send_after(self(), {:maybe_delete_branch, batch.id}, 0)
```
notriddle: Instead of send_after, you can just use send and leave off the 0.

PR #238, opened 2017-07-25
Show the current commit in the footer

### LINKED ISSUE #225: Include latest built Git SHA on the bors frontend
It would be helpful to display the latest Git SHA on the bors fronted application, preferably with a link to the commit/release on Github.  That would it easy for developers to see if their instance of bors-ng is up to date or not.

My thoughts are that it would go [in the footer,](https://github.com/bors-ng/bors-ng/blob/master/apps/bors_frontend/web/templates/layout/app.html.eex#L43) next to the disclaimer about using the free service.  A mockup would look something like this:

**Before**
footerwithoutsha

**After**
footerwithsha

Or something similar.  Inspired by https://github.com/IMA-WorldHealth/bhima-2.X/pull/1859#issuecomment-315174010.

notriddle:
Fixes #225

### DIFF
--- a/apps/bors_frontend/web/views/layout_view.ex
+++ b/apps/bors_frontend/web/views/layout_view.ex
@@ -5,4 +5,43 @@ defmodule BorsNG.LayoutView do
   """
 
   use BorsNG.Web, :view
+
+  def get_version do
+    get_heroku_commit() || get_git_commit() || get_release_version()
+  end
+
+  def get_heroku_commit do
+    "HEROKU_SLUG_COMMIT"
+    |> System.get_env()
+    |> case do
+      nil -> nil
+      commit -> String.slice(commit, 0..10)
+    end
+  end
+
+  def get_git_commit do
+    ".git/HEAD"
+    |> File.read()
+    |> case do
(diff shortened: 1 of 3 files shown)


apps/bors_frontend/web/views/layout_view.ex
```diff
@@ -5,4 +5,43 @@ defmodule BorsNG.LayoutView do
...
+      _ -> nil
+    end
+    |> case do
+      nil -> nil
+      commit -> String.slice(commit, 0..10)
+    end
+  end
+
+  def get_release_version do
+    case :application.get_key(:vsn) do
```
jniles: I'm not completely familiar where this `vsn` comes from.  Is it based on Git versions?  Will it resolve correctly via the GitHub tree URL it is templated into?

notriddle: vsn is the version provided in the Mix.exs file. It's the fallback in case we can't find a git commit anywhere else.

jniles: Ah, I gotcha.  That seems like an acceptable fallback.

PR #239, opened 2017-07-25
Run CI against Elixir 1.5

notriddle:


notriddle:
bors r+

This will need to recompute the dialyzer results, so it'll run out of memory the first time. I'll need to re-try the build, which will fix that.

PR #241, opened 2017-07-27
Cut out another reference to github.io

notriddle:


notriddle:
bors r=jniles

notriddle:
Dialyzer doesn't work when the build cache isn't fully populated, and apparently it's bad enough that it runs out of memory.

I think we're going to need to reduce the number of configurations that it dialyzes. Or even just move dialyzer to an environment where it just gets enough RAM to do a from-scratch run without running out.

jniles:
Weird... Any idea why is it crashing on this particular PR?

notriddle:
I just added another Elixir version to the build matrix, so maybe I'm putting too much in the build cache now and it's running out of space?

notriddle:
bors r=jniles

PR #244, opened 2017-07-28
Use openssl tool to generate the base64 file

notriddle:
This should be portable, and will avoid the strings nonsense.


notriddle:
bors r+

This can't possibly be a legitimate test failure. Apparently, PostgreSQL failed to start.

PR #247, opened 2017-07-28
Add a setting for github app URL

### LINKED ISSUE #246: Add a configuration option for the "App URL"
There are a few places within bors-ng where it hard-codes the bot name "bors". The command names can probably just stay as is, since it makes more sense to say "bors r+" than "bors-voyager r+".

However, the "Add/Remove Repositories" button is just wrong. Right now, it points directly at the public instance of bors's page, when it ought to vary by bot.

Instead of hard-coding "bors" on <https://github.com/bors-ng/bors-ng/blob/master/apps/bors_frontend/config/config.exs#L14>, it should use `"#{System.get_env("GITHUB_APP_PUBLIC_LINK") || "https://github.com/apps/bors"}/installations/new"`

carols10cents:
So that when you add repositories to your own instance of bors-ng,
github doesn't add the main instance of bors-ng instead 😱

Fixes #246.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ The *webhook secret* should be a randomly generated string. The `mix phoenix.gen
 
 GitHub will send a "ping" notification to your webhook endpoint. Since bors is not actually running yet, that will fail. This is expected.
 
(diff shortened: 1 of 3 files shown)


notriddle:
That worked!

PR #249, opened 2017-07-29
Use "merged_at" to determine if a PR is merged

notriddle:
While the "merged" boolean key is included in some webhooks that include a pull request, it is not sent with pull request reviews.

The missing key causes the pull request reviews to produce a 500 error and including a "bors r+" in one would not work.


notriddle:
r+'ed it myself because nobody else was looking at it, and it's causing 500 errors to show up in the logs.

PR #250, opened 2017-07-29
Remove the LICENSE-APACHE link

notriddle:
People who want to click the link to open it can use the apache.org link.

More importantly, while the link worked within GitHub, it doesn't work in ExDoc (because ExDoc doesn't include that plain text file), so it caused a broken link in <https://bors.tech/devdocs/bors-ng/readme.html>.


notriddle:
bors r=jniles

Travis flaked out, I think.

PR #255, opened 2017-08-03
Drop in the Scout agent

notriddle:
Let's stop Heroku from spamming me about monitoring.


notriddle:
bors r+

Since nobody else seems interested.

PR #258, opened 2017-08-06
priority implementation

### LINKED ISSUE #201: Priority
* Make sure that every PR in a batch has the same priority. This is the same way bors-ng does multiple branches.

* Pick PRs with a higher priority over ones with a lower priority. That's an ORDER BY clause.

* When a PR gets added, if the current running batch has a lower priority than the new PR, cancel the current batch and put the batch's PRs back into the queue.

CC https://github.com/servo/homu/issues/102

khodzha:
closes #201

### DIFF
--- a/apps/bors_worker/lib/batcher.ex
+++ b/apps/bors_worker/lib/batcher.ex
@@ -71,6 +71,10 @@ defmodule BorsNG.Worker.Batcher do
   end
 
   def do_handle_cast({:reviewed, patch_id, reviewer}, project_id) do
+    do_handle_cast({:reviewed, patch_id, reviewer, 0}, project_id)
+  end
+
+  def do_handle_cast({:reviewed, patch_id, reviewer, priority}, project_id) do
     case Repo.get(Patch.all(:awaiting_review), patch_id) do
       nil ->
         # Patch exists (otherwise, no ID), but is not awaiting review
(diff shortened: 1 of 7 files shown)


khodzha:
i've only changed command parsing and added p parameter yet, so its early WIP

notriddle:
Write a test case for it, and I'll call it good!

khodzha:
i need to handle cancellation of running batches with lower priorities too

apps/bors_worker/lib/batcher.ex
```diff
@@ -516,4 +545,11 @@ defmodule BorsNG.Worker.Batcher do
   defp get_repo_conn(project) do
     Project.installation_connection(project.repo_xref, Repo)
   end
+
+  defp put_incomplete_on_hold(batch) do
```
khodzha: not sure if im doing right thing here :thinking:

notriddle: Seems decent enough to me.

khodzha:
also should we support bors p=10 syntax?

apps/bors_worker/lib/batcher.ex
notriddle: The biggest missing feature is that it doesn't do anything about the status as shown on the GitHub PR page. It's still going to say that it's "Running," when it should probably change the commit status so that it says it's been put on hold.

khodzha: good point, i should add this

notriddle:
Yes, I think we should support that.

In fact, what I would suggest is that instead of having a priority option on the `activate_by` message, add another message that the Batcher recognizes (`set_priority` maybe?). When `bors p=<num>` is posted, the command runner sends a message to set the priority, and when `bors r+ p=<num>` is posted, the command runner sends the `set_preiority` message followed by the `activate_by` message.

notriddle:
One other thing I noticed missing.

It looks like it supports these two syntaxes:

* `bors p=10` will set the priority.
* `bors r=somebody p=10` will set the priority then start it going with somebody as the reviewer.

It doesn't look like it supports what I think might actually be the most common syntax:

* `bors r+ p=10` should mark it p=10 and put it in the queue.

PR #259, opened 2017-08-07
Block search engines from the admin console

notriddle:
They're just going to get hit with a redirect to the GitHub login page anyway.


jniles:
A dogma test hung briefly, but on restart completed without a hitch.  LGTM!

PR #264, opened 2017-08-11
Remove the status table entries when putting a batch on hold

notriddle:
This commit also adds an integration test for putting a batch on hold and having bors resume it later. Because bors creates status table entries when it starts a batch, it was crashing on a database constraint when it tried to make duplicates.


khodzha:
looks fine to me

notriddle:
bors r= @khodzha

Issue #252, opened 2017-08-01
Post a warning if they try to r+ or try a PR with `[ci skip]` in the title or body


notriddle:
See https://github.com/ordian/toml_edit/pull/3 for an example of how annoying the perpetual timeouts are.

jeffparsons:
This behaviour is irritating, especially in combination with Dependabot.

Dependabot opens a PR with a list of summarised commits from upstream. Some of those commits might have "[ci skip]" in their description.

"CI skip" means absolutely nothing in our build system, but Bors predicts (wrongly) that the presence of "CI skip" in the PR description means that the PR would time out. The PR then sits around until someone manually edits the description to remove the words "CI skip".

PR #265, opened 2017-08-12
skip attempt or batch if ci skip is present in patch description

khodzha:
#252

### DIFF
--- a/apps/bors_database/test/models/patch_test.exs
+++ b/apps/bors_database/test/models/patch_test.exs
@@ -227,4 +227,14 @@ defmodule BorsNG.Database.PatchTest do
     }
     assert Patch.ci_skip?(p)
   end
+
+  test "ci_skip? returns false if ci skip is absent" do
+    p = %Patch{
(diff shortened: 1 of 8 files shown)


khodzha:
not sure how i should test skipping batch/attempt run

khodzha:
i think i accidentally skipped ci :rofl:

apps/bors_database/test/models/patch_test.exs
```diff
@@ -207,4 +207,24 @@ defmodule BorsNG.Database.PatchTest do
...
+
+  test "ci_skip? checks title" do
+    p = %Patch{
+      pr_xref: 9,
+      title: "[ci skip] title",
+      body: "this is body",
+      commit: "C"
+    }
+    assert Patch.ci_skip?(p)
+  end
```
notriddle: This is conspicuously missing a negative test.

notriddle:
Since the PR message itself doesn't contain `[ci skip]`, it should still work when I r+ it.

PR #268, opened 2017-08-15
Improve some look-and-feel stuff in the dashboard

notriddle:
* The top navigation bar is now slightly shorter, making it exactly the same height as the one on bors.tech. [Make My Logo Bigger Cream] is not getting any of my business!
* There is now more space between the different parts of the merge queue (the Running, Awaiting review, Waiting to run sections), so that the header is closer to the items it goes with than the items it doesn't go with.
* A count of the items in each part of the merge queue has been added.
* The priority and commit hash are shown for the patches, in addition to the title and number.
* It now has a favicon.
* The cancel and sync buttons are on the bottom, since they're not used as often.
* A scrollbar is always shown, even if the window is tall enough. This prevents the content from "jumping" when you switch between a page that needs a scrollbar and one that is short enough to not need one.

[make my logo bigger cream]: http://www.makemylogobiggercream.com/


notriddle:
bors r=carols10cents,notriddle

http://logs.glob.uno/?c=mozilla%23rust-infra&s=15+Aug+2017&e=16+Aug+2017#c23754

PR #273, opened 2017-08-17
search user by github id before trying to create

### LINKED ISSUE #272: Adding a github user with improper capitalization to reviewers causes 500 Internal Server Error
On the repository settings page, adding a valid github user with improper capitalization renders a 500 Internal Server Error.  This doesn't seem to work with an invalid github user (e.g. a UUID). However, if the username does not conform to the capitalization found in the URL of github.com, it will throw this error.

Steps to reproduce:
 1. Navigate to the repo settings pages: `https://app.bors.tech/repositories/:id/settings`
 2. Attempt to add a reviewer from a Github profile with their capitalization wrong (such as all caps)
 3. Be routed to a 500 error page.

Expected Behavior:
 1. Flash a relevant error message such as when an invalid user is added
OR
 2. Add the user - it looks like [some other projects](https://github.com/pullapprove/support/issues/35) have implemented case-sensitivity to get around something similar.  Not sure if it is possible or relevant here.

khodzha:
Fixes #272

### DIFF
--- a/apps/bors_frontend/web/controllers/project_controller.ex
+++ b/apps/bors_frontend/web/controllers/project_controller.ex
@@ -155,8 +155,15 @@ defmodule BorsNG.ProjectController do
         |> case do
           nil -> nil
           gh_user ->
-            %User{user_xref: gh_user.id, login: gh_user.login}
-            |> Repo.insert!()
+            case Repo.get_by(User, user_xref: gh_user.id) do
+              nil ->
+                User.changeset(%User{}, %{
+                  user_xref: gh_user.id,
+                  login: gh_user.login
+                })
+                |> Repo.insert!()
+              user -> user
+            end
         end
       user -> user
     end
--- a/apps/bors_database/lib/user.ex
+++ b/apps/bors_database/lib/user.ex
@@ -39,5 +39,6 @@ defmodule BorsNG.Database.User do
   def changeset(struct, params \\ %{}) do
     struct
     |> cast(params, [:user_xref, :login, :is_admin])
+    |> unique_constraint(:user_xref, name: :users_user_xref_index)
   end
 end
--- a/apps/bors_database/priv/repo/migrations/20170823012140_change_users_login_type_to_citext.exs
+++ b/apps/bors_database/priv/repo/migrations/20170823012140_change_users_login_type_to_citext.exs
@@ -1,0 +1,19 @@
+defmodule BorsNG.Database.Repo.Migrations.ChangeUsersLoginTypeToCitext do
+  use Ecto.Migration
+
+  def up do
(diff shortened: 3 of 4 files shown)


khodzha:
i will add test for this, dont merge it yet

notriddle:
You should be able to add the S-do-not-merge-yet tag yourself, right?

notriddle:
Also, I think there's a way to have a case-insensitive column in PostgreSQL. Maybe that would be a simpler way to do it?

notriddle:
It looks like you're able to edit labels now. 😆 

Also, [apparently, Ecto does support case-insensitive columns](https://elixirforum.com/t/case-insensitive-column-in-ecto/2062), though it gets kind of complex with either custom indexes or extensions. Making it work across both PostgreSQL and MySQL could get kinda complex.

It should be doable, though. Actually, based in Ecto's description of the setup, it might already be case-insensitive on MySQL, so you just need to make sure your migration only runs on Postgres.

khodzha:
im not sure if we should rely on extension, should we store logins downcased? or create separate column for downcased login

khodzha:
i will try citext for now

notriddle:
AFAICT, [citext is a first-party extension](https://www.postgresql.org/docs/9.1/static/citext.html). It's in the Portable PostgreSQL package that I use on Windows by default, it's in Heroku PostgreSQL by default, and it's in Ubuntu's `postgresql-contrib` package.

notriddle:
@khodzha Any progress on this one? Or any questions, need help?

khodzha:
ok, im not sure if i did migration properly and if it is "postgres only" migration

notriddle:
I just found out that the CI for checking MySQL is broken. That's nice. #280

apps/bors_frontend/web/controllers/project_controller.ex
```diff
@@ -155,8 +155,15 @@ defmodule BorsNG.ProjectController do
...
-            |> Repo.insert!()
+            case Repo.get_by(User, user_xref: gh_user.id) do
+              nil ->
+                User.changeset(%User{}, %{
+                  user_xref: gh_user.id,
+                  login: gh_user.login
+                })
+                |> Repo.insert!()
+              user -> user
+            end
```
notriddle: These changes shouldn't be needed any more.

notriddle:
https://github.com/khodzha/bors-ng/pull/1

apps/bors_frontend/web/controllers/project_controller.ex
khodzha: well i guess so...
but it doesnt hurt. or does it? :thinking:

PR #277, opened 2017-08-21
Link to the forum instead of the IRC

notriddle:
For the long-term, it's probably better that support is done in a way that's more accessible to search engines and has actual *threading*.


khodzha:
maybe add link to forum in readme too?

notriddle:
The forum is now in the README.

notriddle:
bors r=khodzha

PR #284, opened 2017-08-25
Mention the Internal/External App in the README

### LINKED ISSUE #283: Restricting repository add/remove
I think the Rust organization would probably limit its use to their own repositories, so I think we may need an option to make it admin-only or more configurable one.

notriddle:
Closes #283

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,11 +188,11 @@ The first step is to [register a new Github App] on the GitHub web site.
 
 The *Github App name*, *description*, and *homepage URL* are irrelevant, though I suggest pointing the homepage at the dashboard page.
 
-The *user authorization callback URL* should be at `<dashboard url>/auth/github/callback`.
+The *user authorization callback URL* should be at `<<DASHBOARD URL>>/auth/github/callback`.
 
 Leave the *setup URL* blank.
 
-The *webhook URL* should be at `<dashboard page>/webhook/github`.
+The *webhook URL* should be at `<<DASHBOARD URL>>/webhook/github`.
 
 The *webhook secret* should be a randomly generated string. The `mix phoenix.gen.secret` command will work awesomely for this. Keep this handy to specify the same value in the bors configuration (you can also edit this value later if you need to).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -242,9 +242,23 @@ You'll need the following values from your GitHub App for configuring bors-ng:
 
 - Private key (generate one and download the file)
 - OAuth credentials
-- Public link (looks like `https://github.com/apps/your-app-name`; should be a convenient "copy to clipboard" button next to it in the right hand column)
+- Public link (looks like `https://github.com/apps/<<APP NAME>>`; should be a convenient "copy to clipboard" button next to it in the right hand column)
 - ID (appears beneath the Public link and "Owned by" in the right hand column)
 
+### Internal app?
+
+GitHub Apps can be set as "Internal" or "External."
```
ishitatsuyuki: Quotes mismatch here, put the period out.

README.md
```diff
@@ -188,11 +188,11 @@ The first step is to [register a new Github App] on the GitHub web site.
 
 The *Github App name*, *description*, and *homepage URL* are irrelevant, though I suggest pointing the homepage at the dashboard page.
 
-The *user authorization callback URL* should be at `<dashboard url>/auth/github/callback`.
+The *user authorization callback URL* should be at `<<DASHBOARD URL>>/auth/github/callback`.
```
ishitatsuyuki: I don't think it needs to be double quote and shouting.

notriddle: I changed this after realizing that I was using a different interpolation syntax in different parts of the README (notice the spot where `https://github.com/apps/your-app-name` is changed to `https://github.com/apps/<<APP NAME>>`).

The reason for going with shouting and double-quote, of course, is so that it doesn't look like a part of the URL. I figure there's no real harm in being as unsubtle as you can short of doing real harm (two quotes are fine, two hundred would break the text layout).

README.md
```diff
@@ -242,9 +242,23 @@ You'll need the following values from your GitHub App for configuring bors-ng:
 
 - Private key (generate one and download the file)
 - OAuth credentials
-- Public link (looks like `https://github.com/apps/your-app-name`; should be a convenient "copy to clipboard" button next to it in the right hand column)
+- Public link (looks like `https://github.com/apps/<<APP NAME>>`; should be a convenient "copy to clipboard" button next to it in the right hand column)
 - ID (appears beneath the Public link and "Owned by" in the right hand column)
 
+### Internal app?
+
+GitHub Apps can be set as "Internal" or "External".
```
notriddle: Fixed

README.md
```diff
@@ -188,11 +188,11 @@ The first step is to [register a new Github App] on the GitHub web site.
 
 The *Github App name*, *description*, and *homepage URL* are irrelevant, though I suggest pointing the homepage at the dashboard page.
 
-The *user authorization callback URL* should be at `<dashboard url>/auth/github/callback`.
+The *user authorization callback URL* should be at `<DASHBOARD URL>/auth/github/callback`.
```
notriddle: Done.

PR #285, opened 2017-08-26
Synchronize the commit hash during regular patch synchronization

### LINKED ISSUE #217: Patch sync should update the commit, not just the open state
Currently, bors-ng assumes that if the commit hash of a PR changes, then a "synchronize" event will be delivered. Similarly, it assumes that "edited" events will be delivered if a PR body or title is changed. The manual sync [only updates the "open" flag](https://github.com/bors-ng/bors-ng/blob/08c369152d223e22ba6b0ec9102c1b2c59de8918/apps/bors_worker/lib/syncer.ex#L81).

These aren't perfectly reliable. Besides the fact that webhooks can just get dropped, "synchronize" events aren't delivered if the head of a PR gets its commit changed while the PR is closed (as demonstrated in slide-rs/specs#198).

notriddle:
Fixes #217

### DIFF
--- a/apps/bors_frontend/test/controllers/webhook_controller_test.exs
+++ b/apps/bors_frontend/test/controllers/webhook_controller_test.exs
@@ -62,6 +62,44 @@ defmodule BorsNG.WebhookControllerTest do
     assert "OTHER_BRANCH" == patch2.into_branch
   end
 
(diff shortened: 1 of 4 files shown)


Macarse:
👍

notriddle:
bors r= @Macarse, @khodzha

PR #290, opened 2017-09-04
use _path url helpers instead of _url in layout template

khodzha:
there's no need to use full urls in these links

though actual reason behind this change is i've set my own test instance, configured bors_frontend.url with host public ip, yet links are displayed as `https://admin`

### DIFF
--- a/apps/bors_frontend/web/templates/layout/app.html.eex
+++ b/apps/bors_frontend/web/templates/layout/app.html.eex
@@ -21,10 +21,10 @@
   <body>
     <header role="banner">
       <div class="wrapper wrapper--mini">
-        <a id=header-logo href="<%= page_url(@conn, :index) %>"><img alt="bors" src='<%= static_path(@conn, "/images/a.svg") %>' width="90" height="25"></a>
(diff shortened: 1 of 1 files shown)


Macarse:
Is this fixing a bug?

khodzha:
i havent investigated source of current behaviour, so maybe i did a mistake in configuration, cant say for sure

Macarse:
@khodzha ok! The change makes sense but I would wait for @notriddle 👍

notriddle:
It's a configuration problem on your end. You need to fix:

https://github.com/bors-ng/bors-ng/blob/master/apps/bors_frontend/config/prod.secret.exs#L5

But shaving off bytes is cool.

bors r+

notriddle:
Maybe you could post this file in a Gist somewhere, see if I can figure out what's wrong?

khodzha:
ah, i guess i know what's the problem. currently it's
```url: [host: {:system, "207.154.195.219"}, scheme: "https", port: 443],```
and i think it should be
```url: [host: "207.154.195.219", scheme: "https", port: 443],```
right?

khodzha:
yeah, checked it and i am right

PR #297, opened 2017-09-20
Add TaskCluster CI configuration

notriddle:
[TaskCluster] is basically "hard mode" Travis CI. In particular, it supports things like "asking for a particular worker type" (because a certain type of task needs more RAM or whatever), triggering builds from things other than pushing to a branch or opening a PR, and having completely different tasks for pull requests and branches (you'll notice we do so).

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,0 +1,88 @@
+version: 0
+allowPullRequests: public
+tasks:
+
+  # The bors build queue (test suite)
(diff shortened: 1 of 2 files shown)


notriddle:
Because of the way it works, we can't really start gating on TaskCluster until after the YML file is in master. If we do decide to go ahead with this, a separate PR will add TaskCluster to the bors.toml file.

Issue #23, opened 2017-01-09
Show a "nothing to see here" placeholder on the dashboard home page if there are no patches awaiting review


notriddle:
Exactly what it sounds like. It should say that nothing is ready to review, instead of showing the table headers followed by nothing at all.

notriddle:
It's defined in [dashboard.html.ex](https://github.com/bors-ng/bors-ng/blob/master/apps/bors_frontend/web/templates/page/dashboard.html.eex).

Issue #22, opened 2017-01-09
Show a "nothing to see here" placeholder on the patch list for projects that have no open patches


notriddle:
Exactly what it sounds like. You should say that there are no open pull requests, instead of just not showing anything.

notriddle:
It's in [project/index.html.eex](https://github.com/bors-ng/bors-ng/blob/master/apps/bors_frontend/web/templates/project/index.html.eex).

abeltensor:
I have a solution if you are interested.  

I implemented two helper functions into the project_view.eex folder to check if the projects list is empty. Then I added a few lines into the index.eex file which uses these new functions to show the table or show the "nothing to see here" placeholder.  Its a fairly basic solution but I figured you would want to build on it from there. 

heres what index would look like:

`<nav><div class=wrapper>
<h1>Repositories</h1>
</div></nav>

<main role=main><div class=wrapper>

<div class=toolbar>
  <a class=toolbar-item href="<%= BorsNG.Attrs.get_github_integration_url() %>">
    Add / Remove Repositories
  </a>
</div>
<%= if empty? @projects do %>
  Nothing to See here!
  <% else %>

<table class="table">
  <tbody>
<%= for project <- @projects do %>
    <tr role=row>
      <td class="fill-link">
        <%= link project.name, to: project_path(@conn, :show, project) %>
      </td>
    </tr>
<% end %>
<% end %>
  </tbody>
</table>

</div></main>`

Here are the empty? helper functions:
`def empty?([]), do: true
  def empty?(list) when is_list(list) do
    false
  end
`

notriddle:
Is there some reason you didn't just use `if @projects == []`?

And, yes, I'm interested! Thanks for working on this :+1:

abeltensor:
I used the helper functions because it just seemed like a more elegant solution then hard coding the conditional.  Obviously either or still works however.

notriddle:
That makes sense, let's go with that then.

abeltensor:
No problem, shall I make a pull request then?

notriddle:
Yep. :octocat:

PR #302, opened 2017-10-03
empty? helper functions for #22 and #23

abeltensor:
Added two helpers to the page_view.ex and project_view.ex views and then implemented these functions to solve issues #22 and #23.

Please check my code, was unable to rebuild the full project on my localhost the 2nd time
(should work fine though).

### DIFF
--- a/apps/bors_frontend/web/templates/page/dashboard.html.eex
+++ b/apps/bors_frontend/web/templates/page/dashboard.html.eex
@@ -7,6 +7,9 @@
     <tr role=row><th>#</th><th>Title</th><th>Repository</th></tr>
   </thead>
   <tbody>
+<%= if empty? @patches do %>
+  Nothing to see here
+<%= else %>
 <%= for patch <- @patches do %>
     <tr role=row>
       <td>
@@ -21,6 +24,7 @@
         <%= link patch.project.name, to: project_path(@conn, :show, patch.project) %>
       </td>
     </tr>
+<% end %>
 <% end %>
   </tbody>
 </table>
--- a/apps/bors_frontend/web/templates/project/index.html.eex
+++ b/apps/bors_frontend/web/templates/project/index.html.eex
@@ -9,6 +9,9 @@
     Add / Remove Repositories
   </a>
 </div>
+  <%= if empty? @projects do %>
+    Nothing to see here!
+  <% else %>
 
 <table class="table">
   <tbody>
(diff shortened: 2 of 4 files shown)


notriddle:
I'm sorry to inform you that [your code is not valid](https://travis-ci.org/bors-ng/bors-ng/jobs/282833235). Your template tries to call `is_empty?`, but you actually named the function `empty?`.

notriddle:
I'm not sure if you realized this, but these two function definitions are equivalent:

```elixir
# Both of these are valid Elixir

# Notice how the single-line version has no `end` keyword, but does have a comma (`,`) and a colon (`:`)
fn empty?(list) when is_list(list), do: false

# Notice how the multi-line version has no comma (`,`) or colon (`:`), but does have an `end` keyword
fn empty?(list) when is_list(list) do
  false
end
```

You wrote something like this, which is not valid.

```elixir
# This is not valid Elixir. The single-line version has no `end` keyword.
fn empty?(list) when is_list(list) do false end
```

https://hexdocs.pm/elixir/1.5/syntax-reference.html is really useful for this stuff.

abeltensor:
Whitespace should be completely gone now.

abeltensor:
Looks like the travis ci build is finally going to work.  

Problem with whitespace came from using the github editor.  I figured such a small change would be more easily implemented this way (I stand corrected).  

Problem with function name was simply my own negligence.  

Hopefully everything else works out as intended; this is technically my first collaboration so I hope you will forgive the hiccups.

notriddle:
No problem, it's not like your actually hurting anything, 🤷‍

Besides, the whole point of having Easy Issues lying around is so that people can learn how to perform all the ceremony without simultaneously having to diagnose complicate component interaction puzzles. 😖

So, mission accomplished.

bors r+

abeltensor:
Thank you.  Yes it is a learning experience.  At least I know I should probably just push everything from emacs next time instead of using the github editor haha.

notriddle:
By the way, about:

> Please check my code, was unable to rebuild the full project on my localhost the 2nd time (should work fine though).

What kind of problem was preventing you from building it locally? Was it just the same errors that Travis gave, or were you AFK, or something else?

abeltensor:
so i am building on windows, the error I got had to do with one of the private functions in the router not being able to find a user.  

When I built it last night, there were not problems like that; I try rebuilding it again and see if there is a problem.  Its very likely that I just didn't install all of the node dependencies properly.

PR #303, opened 2017-10-03
Add the placeholder to the repository page

notriddle:
Probably should've mentioned it in one of those issues, too, @abeltensor


abeltensor:
Ill try rebuilding the current build and see if I get the same error.  Then If I do, ill submit an issue regarding it.

Issue #308, opened 2017-10-06
bors does not respond to "Bors"


jniles:
This may be intended behavior, but I wanted to bring up for discussion.

On mobile browsers, my keyboard aggressively autocorrects every first line in Github's text area and capitalizes it.  This leads to commands such as "Bors r+", which seem to be unrecognized.  It will respond to a lowercase "b".

I can see how this might be useful to make sure bors doesn't do any unintended behavior and so [I have added it to the wiki](https://forum.bors.tech/t/quirks-of-how-bors-ng-parses-github-pr-comment-commands/53).  However, if it is simply a corner case, it would be nice to have it fixed!

notriddle:
I did that on purpose so that you can avoid triggering it by using sentence case, but you're right that it makes it harder to use when your keyboard wants you to type like that...

If nobody else has an opinion on how it ought to work, I'm fine with switching to a case-insensitive match for "bors".

notriddle:
Also, we probably have a sufficiently large set of ways to avoid accidentally triggering it...

GustavoCaso:
@notriddle I love to help with this one.

jniles:
@GustavoCaso, I don't believe anyone is working on it.  Go for it!

PR #318, opened 2017-10-17
Allow Command to parse Bors and bors

GustavoCaso:
#308 

I decided to use `Regex` in this case because it seems the most simple way to keep functionality without changing too much code.

@notriddle please add as many comments as need it.

### DIFF
--- a/apps/bors_frontend/web/command.ex
+++ b/apps/bors_frontend/web/command.ex
@@ -107,20 +107,19 @@ defmodule BorsNG.Command do
   def parse(comment) do
     comment
     |> String.splitter("\n")
-    |> Enum.flat_map(fn
-      @command_trigger <> " " <> cmd ->
-        trim_and_parse_cmd(cmd)
-      @command_trigger <> ": " <> cmd ->
-        trim_and_parse_cmd(cmd)
-      _ -> []
+    |> Enum.flat_map(fn(string) ->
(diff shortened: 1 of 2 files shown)


apps/bors_frontend/web/command.ex
```diff
@@ -107,20 +107,19 @@ defmodule BorsNG.Command do
...
-      @command_trigger <> ": " <> cmd ->
-        trim_and_parse_cmd(cmd)
-      _ -> []
+    |> Enum.flat_map(fn(string) ->
+      trim_and_parse_cmd(Regex.named_captures(regex(), string))
     end)
   end
 
-  def trim_and_parse_cmd(cmd) do
+  def regex, do: ~r/^#{@command_trigger}:?\s(?<command>.+)/i
```
Macarse: elixir doesn't offer a way to compile regex, right?

notriddle: I think [this *is* the way to compile regex](https://hexdocs.pm/elixir/Regex.html).

> Regular expressions created via sigils are pre-compiled and stored in the `.beam` file.

notriddle:
Just as a quick tip, @GustavoCaso, if you include `closes #308`, instead of just `#308`, in your issue number, then the issue will automatically be closed as soon as your PR gets merged.

GustavoCaso:
@notriddle awesome thanks for the quick tip

PR #319, opened 2017-10-27
Get the same navigation bar from the home page onto the app

notriddle:
This also makes the Repositories screen more accessible, by surfacing it out of the menu (on screen where that fits).

wide-mode

narrow-mode


notriddle:
bors r+

No complaints, plus an approval, so going ahead!

Issue #251, opened 2017-07-30
Add "retry" and a list of people with only retry, not r+ permissions


carols10cents:
In rust-lang/rust, we have flaky builds sometimes 😰

We also have an infrastructure team that has a rotation of PR triage duty. One of the things PR triagers do is retry bors/homu if something intermittent has caused a build to fail.

There are some folks who have permissions to ask bors to retry, but they don't have general r+ permissions. [These permissions actually cover try, retry, force, clean, and prioritization](https://github.com/rust-lang/rust-central-station/blob/f52d7bbce81f438b227535117d0f437eb2e23b66/homu.toml.template#L92).

This lets more people help keep PRs moving but without them needing full r+ permissions.

PR #321, opened 2017-11-02
Add a second, less-permissive mode for users to be in

notriddle:
Part of #251 

Mere "members," as opposed to "reviewers," have permission to run "bors try" and will have permission to run "bors retry" once that's also implemented.


Macarse:
bors +r

Macarse:
@notriddle LGTM

PR #322, opened 2017-11-06
Use full-sized commit hash in the footer link

notriddle:


apps/bors_frontend/web/views/layout_view.ex
```diff
@@ -7,19 +7,27 @@ defmodule BorsNG.LayoutView do
...
-    {String.slice(hash, 0..6), hash}
+    get_commit() || get_tag()
   end
 
-  def get_heroku_commit do
-    "HEROKU_SLUG_COMMIT"
-    |> System.get_env()
-    |> case do
-      nil -> nil
-      commit -> String.slice(commit, 0..10)
```
Macarse: why it changed from `0..10` to `0..6`?

notriddle: That's how many characters GitHub itself shows.

PR #328, opened 2017-12-13
Allow commands in the PR body

notriddle:
This allows the user to include a priority declaration in it, or to do the common-justification-plus-self-r+ in it.


notriddle:
bors ping

notriddle:
Failed to connect to GitHub...

I'll just leave it be for a few hours until Heroku goes back up, and maybe everybody else will get working again, too.

PR #329, opened 2018-01-02
Use Confex to load config variables at runtime

notriddle:
This should make OTP releases more useful, by allowing you to compile bors once and deploy the same image multiple times.

https://forum.bors.tech/t/deploying-bors-ng-in-production-using-docker-building-a-release-seems-to-hardcode-configuration/97


notriddle:
Spurious failure to contact hex.pm.

bors r+

PR #333, opened 2018-01-06
Implement fixes for release-based container deployments

danielkza:
Fix some configuration options that made production container deployments difficult, and add a custom module that can do the same job as `mix ecto.create` and `mix ecto.migrate`, as Mix tasks cannot be run from a compiled releases (based on the [sample code](https://github.com/bitwalker/distillery/blob/master/docs/Running%20Migrations.md) from the Distillery docs).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+_build
+deps
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,53 @@
+ARG ELIXIR_VERSION=1.4.5
+
+FROM elixir:${ELIXIR_VERSION} as builder
+
+ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
+
+RUN apt-get update -q && apt-get install -y build-essential libtool autoconf curl
+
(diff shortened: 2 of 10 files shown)


danielkza:
There is one detail that I don't know how to address correctly: Ecto does not parse the `ssl` option from the database URL in currently available releases (only [in master](1178f41be9bef3f2819adfbe8cfa81809fb1555f) as of now. But forcing SSL to be on makes it quite a bit more difficult to test a built production container with a linked DB container. Not forcing it will probably break Heroku deployments by default.

Does parsing the URL manually make sense, or should a new environment variable be introduced, while keeping SSL enabled by default in prod?

notriddle:
Personally, I'd just go ahead and point at the git commit in mix.exs, instead of working around it. It's not like we're pushing bors-ng itself to Hex.

notriddle:
bors r+

Both versions are actually correct. It's just Travis CI has been timing out a lot lately. I need to put in the time to migrate over to TaskCluster completely.

notriddle:
Okay, thanks!

bors r+

notriddle:
Next. Week.

bors r+

PR #338, opened 2018-01-18
log: sort by update time

zimbatm:
the logs where originally sorted by create time but displayed by update
time which was confusing


zimbatm:
How many times are batches updated? To build a full timeline each batch should be represented on creation and on each update.

notriddle:
A batch gets updated every time its state changes, basically.

zimbatm:
So I guess that intermediate states are being lost on update since it's using the relational database and probably a row to store the data

PR #355, opened 2018-02-24
Add red flags for patches associated with multiple batches

### LINKED ISSUE #65: Add a "red flag" for patches that are associated with multiple incomplete batches
If there are multiple waiting-or-running batches with the same patch in them, it's a sign that something has gone wrong, because a patch should not be queued up multiple times. This should be added along with the orphaned-project check that is already there.

kimsaehun:
Fixes #65 

I'm unsure if I implemented the red flag condition correctly but I think I got it right.

Also, I wasn't exactly sure what the "dup-patches.html" should show so I just had it list the duplicate patches. If there's something else that would be better to present I'd like to work on that as well. Got to learn more about Phoenix and would love to to that to more practice.


notriddle:
Everything looks good to me, though now it needs a rebase (sorry). You want to do it, or should I?

kimsaehun:
I'd like to give it a shot! I'll try to get it done within 3 days from now.

kimsaehun:
I believe I rebased correctly seeing as how GitHub states that my branch is "2 commits ahead of bors-ng:master." I've added an extra commit since the template file I created didn't automatically move itself to the new template directory under /lib. Hopefully this works!

notriddle:
Looks good to me!

bors r+

PR #358, opened 2018-02-26
Link to the specific batch ID in commit statuses

grahamc:
- Add batch IDs log rows
 - Set the background color on `<tr>`s targeted via the URL fragment
 - Clean up some trailing whitespace
 - Append the batch ID to send_status

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -127,10 +127,10 @@ defmodule BorsNG.Worker.Batcher do
                 reviewer: reviewer}
               Repo.insert!(LinkPatchBatch.changeset(%LinkPatchBatch{}, params))
               poll_after_delay(project)
-              send_status(repo_conn, [patch], :waiting)
+              send_status(repo_conn, batch.id, [patch], :waiting)
             {:error, message} ->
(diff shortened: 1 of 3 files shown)


lib/worker/batcher.ex
```diff
@@ -550,26 +550,26 @@ defmodule BorsNG.Worker.Batcher do
...
+    send_status(repo_conn, id, patches, message)
     unless is_nil commit do
       {msg, status} = Batcher.Message.generate_status(message)
       repo_conn
       |> GitHub.post_commit_status!({
         commit,
         status,
         msg,
-        project_url(Endpoint, :log, project_id)})
+        project_url(Endpoint, :log, project_id) <> "#batch-#{id}"})
```
grahamc: afaict there is no way to put attach a URL fragment inside the URL helper itself.

notriddle: I'm not too worried about this. We know that the batch ID is a self-generated integer, so it's safe to stick it into a URL (if it were computed from user input, I'd be more worried).

grahamc:
P.S. I'm happy to write tests for things, too. The only hangup is I'm quite novice at Elixir though, so I'd probably need a little bit of direction.

Issue #313, opened 2017-10-13
delete old history


notriddle:
After about six months or so,

* A batch that is marked succeeded/failed/canceled should be deleted.
* A patch that is marked closed and/or is associated with a batch that is marked succeeded/failed/canceled should be deleted.

If people want statistics, we could dump stuff to something like S3. The point is to get the data out of high $/GiB PostgreSQL storage.

grahamc:
So this seems to be doing the business:

```elixir
# Delete all batches older than N months which are done processing
BorsNG.Database.Repo.delete_all(from p in BorsNG.Database.Batch, where: (p.state in [ ^(:ok)] and p.updated_at < datetime_add(^NaiveDateTime.utc_now, -6, "month") ))
# Delete all the patches which are not open, and older than N months
BorsNG.Database.Repo.delete_all(from p in BorsNG.Database.Patch, where: (p.open == false and p.updated_at < datetime_add(^NaiveDateTime.utc_now, -6, "month")))
```

A third option could be deleting any old patch (older than 6mo) with no related batch. What do you think? I'm sort of inclined to just make this a `mix` task that people can call as needed, and possibly make the # of months a parameter.

notriddle:
Yeah, that's fine. Make it a mix task, and make it accessible [the same way the migrations are](https://github.com/bors-ng/bors-ng/blob/105a3f35df95d1264b38df2c6f4ccbf0ffa9e2ef/lib/database/migrate.ex).

grahamc:
This can be closed now

PR #359, opened 2018-03-05
Create bors.cleanup task

grahamc:
I don't quite follow the meaning of:

> make it accessible like migrations are, so that the people using it in Docker and the people doing it in Heroku can both run it

but I was able to run this on heroku okay.

Implements https://github.com/bors-ng/bors-ng/issues/313

### DIFF
--- a/lib/mix/bors.cleanup.ex
+++ b/lib/mix/bors.cleanup.ex
@@ -1,0 +1,62 @@
+defmodule Mix.Tasks.Bors.Cleanup do
+  use Mix.Task
+  import Mix.Ecto
+  import Ecto.Query
+
+
+  @shortdoc "Prune old batches and patches"
+
+  @moduledoc """
+  Delete old batches and patches from the database.
+
+  Old patches and batches which are done processing generally are not
+  useful to keep around. Deleting them regularly can keep the rows of
+  your database tables under the small limits of some database
+  providers.
+
+  ## Examples
+      mix bors.cleanup --months 6
+
+  ## Command line options
+    * `--months` - the integer number of months to keep data
+  """
+
+  @doc false
+  def run(args) do
+    {opts, _, _} = OptionParser.parse args, switches: [months: :integer]
+    case opts do
+      [] ->
+        Mix.shell.info(
+          "Pass `--months <integer>' for the number of months to delete"
+        )
+
+      [ months: months ] ->
+        Mix.shell.info "Cleaning data older than #{inspect months} months"
+        negative_months = 0 - months
+
+        Enum.each repos(), fn(repo) ->
+          ensure_repo(repo, args)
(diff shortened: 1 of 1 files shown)


notriddle:
CC @danielkza

danielkza:
@grahamc The production Docker image is created by building an OTP release with Distillery, which does not include Mix itself, only the compiled Bors code. I don't think Heroku does the same, but just runs directly from the source instead.

In your current version, the task would not be accessible in the compiled release. Unfortunately I'm far from an Elixir expert, and couldn't find an easy way to share code from our project (Bors) and include it in a Mix task. :(

notriddle:
If that's true, then I guess we can just merge this as-is.

One more thing; could you rebase this onto master? It's still putting the file into apps/bors_database, which doesn't exist any more.

grahamc:
Gotcha. Hmm... I wonder if it would be better to use however the dev version is built, so it does include mix? Surely bors doesn't strictly require whatever performance benefits are had from a "production" build.

I rebased, @notriddle. Thanks!

grahamc:
Prior to merging, I'd like to test this on one more deployment.

danielkza:
@grahamc The point of using a compiled release with multi-stage Dockerfile is to reduce the size of the container image. IIRC including all the Elixir and Node dependencies adds hundreds of MBs to the final size.

grahamc:
```
[nix-shell]$ psql bors_dev
psql (9.6.8)
Type "help" for help.

bors_dev=# select count(1) from patches;
 count 
-------
   135
(1 row)

bors_dev=# select count(1) from batches;
 count 
-------
   101
(1 row)

[nix-shell]$ MIX_ENV=dev mix bors.cleanup  --months 1
Compiling 1 file (.ex)
Cleaning data older than 1 months
[debug] QUERY OK source="batches" db=13.5ms queue=16.8ms
DELETE FROM "batches" AS b0 WHERE (b0."state" IN ($1,$2,$3) AND (b0."updated_at" < ($4::timestamp + ($5::numeric * interval '1 month'))::timestamp)) [2, 3, 4, {{2018, 3, 6}, {18, 27, 43, 482559}}, #Decimal<-1>]
[debug] QUERY OK source="patches" db=11.4ms
DELETE FROM "patches" AS p0 WHERE (NOT (p0."open") AND (p0."updated_at" < ($1::timestamp + ($2::numeric * interval '1 month'))::timestamp)) [{{2018, 3, 6}, {18, 27, 43, 560249}}, #Decimal<-1>]

[nix-shell]$ psql bors_dev
psql (9.6.8)
Type "help" for help.

bors_dev=# select count(1) from patches;
 count 
-------
    78
(1 row)

bors_dev=# select count(1) from batches;
 count 
-------
    43
(1 row)
```

LGTM

PR #362, opened 2018-03-12
Fail if changes are requested or not enough people approved the PR

### LINKED ISSUE #357: bors should (optionally?) check for a review status on a PR before batching
Depending on the policy of a repository and business logic, it may be that reviews are required before merging.  In that case, at least optionally, `bors-ng` should be configurable not to accept `bors r+` before the reviewer has approved.  In the case that a person adds `bors r+` before reviewer(s) have approved, it is possible and not unlikely that a PR will not be approved before running CI and then bors will reject the PR at CI completion time.  Instead (at least optionally), bors should not accept an `r+` job.  `bors try` jobs should, of course, not care about this.

grahamc:
Fixes #357

### DIFF
--- a/lib/github/github.ex
+++ b/lib/github/github.ex
@@ -113,6 +113,17 @@ defmodule BorsNG.GitHub do
     labels
   end
 
+  @spec get_reviews!(tconn, integer | bitstring) :: %{
+    "APPROVED": integer,
+    "CHANGES_REQUESTED": integer
+  }
+  def get_reviews!(repo_conn, issue_xref) do
+    {:ok, labels} = GenServer.call(
+      BorsNG.GitHub,
+      {:get_reviews, repo_conn, {issue_xref}})
+    labels
+  end
+
   @spec get_file!(tconn, binary, binary) :: binary | nil
   def get_file!(repo_conn, branch, path) do
     {:ok, file} = GenServer.call(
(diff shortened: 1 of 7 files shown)


grahamc:
Demo: https://github.com/tweag/bors-test/pull/23

notriddle:
Would you, or someone else, like to add this to the [reference docs]?

[reference docs]: https://bors.tech/documentation/

grahamc:
On it!

notriddle:
You might also want to [let JensRantil know](https://forum.bors.tech/t/require-four-eyes/134) that you implemented the feature he wanted. 😁

heinrich5991:
@notriddle I think this is the culprit for https://github.com/ddnet/ddnet/pull/1079#issuecomment-372496230.

heinrich5991:
Maybe the default for number of code reviews should be 0, since the person writing "bors r+" presumably did one.

notriddle:
That's not the bug, actually. The problem is that it's still counting the old Change Request, even though the same user later posted an Approval.

heinrich5991:
AFAICT this is the standard way to deal with Change Requests. Github displays an "Approve Changes" link on the Change Request if you're the author which does what you can see in the PR.

As far as the Github UI is concerned, there's only "1 Approving Review" now.

notriddle:
Yes, that's the bug.

JensRantil:
FYI, I didn't know about this work so I put together https://github.com/tink-ab/four-eyes which solves the same case.

PR #363, opened 2018-03-12
🤦‍♀️

notriddle:
Fix a regression introduced by #361


grahamc:
Was just on my way over to fix this :) thank you!

Issue #377, opened 2018-03-22
Customize the header


notriddle:
The user should be able to add an env variable (`DASHBOARD_HEADER_HTML`)  and it'll insert it in place of the current "Home | Forum | Docs | Dashboard" navigation bar at the top of the dashboard pages. It's basically the same as things like the GITHUB_INTEGRATION_ID environment variable, just gets used differently.

For backwards-compatibility, it should fall back if the env var is missing.

PR #387, opened 2018-03-29
Enable custom dashboard header & footer

### LINKED ISSUE #376: Customize the footer
The user should be able to add an env variable (`DASHBOARD_FOOTER_HTML`) and it'll insert it in place of the current "This service is provided for free on a best-effort basis" disclaimer at the bottom of the dashboard pages. It's basically the same as things like the `GITHUB_INTEGRATION_ID` environment variable, just gets used differently.

For backwards-compatibility, it should fall back if the env var is missing.

couchand:
These changes allow a user to set `DASHBOARD_HEADER_HTML` and `DASHBOARD_FOOTER_HTML` to customize the dashboard.

Fixes #376 #377 

One question I had was about the proper indentation for the default values.  As it is they're just verbatim moved from the template file, but I could see reformatting them.

(My first Elixir change ever!  Figured this would be a good way to get started contributing here, before diving into some of the more involved issues I've got my eye on.)

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -37,6 +37,9 @@ config :bors, BorsNG,
           <a class=header-link href="https://forum.bors.tech">Forum</a>
           <a class=header-link href="https://bors.tech/documentation/getting-started/">Docs</a>
           <b class=header-link>Dashboard</b>
+    """},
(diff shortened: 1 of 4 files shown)


notriddle:
I've just gone ahead and turned the style checker off in the config files. That might not be the greatest thing ever, but I'm more concerned about coding style in the code itself anyway.

Issue #391, opened 2018-04-04
Sort the patches by PR number before merging them


notriddle:
We really want the parent list and commit message to be in the same order.

https://forum.bors.tech/t/ordering-of-commit-parents-pr-numbers/162

PR #392, opened 2018-04-05
Add co-author trailers to merge commit.

couchand:
When creating the merge commit for a batch, add a "Co-authored-by" trailer to the commit message for each author contributing to the batch.

Fixes: #382

@notriddle How do you usually test things like this against GitHub?  I haven't actually validated any of the code in `lib/github/github/server.ex`.

PR #393, opened 2018-04-06
Sort patches before creating merge commit

couchand:
Previously, only the commit message was sorted, but not the
commit parents.  This moves the sort earlier, so that all
processing on the batch happens in the same sorted order.

Fixes: #391 

I removed the previous test that the commit message generation sorted the PRs, since it's no longer valid.  I'd like to add a test that the sorting is still being done, but I'm not really sure where to put it.  The tests in `test/batcher/batcher_test.ex` don't seem to have a clear way to assert on the generated commit.  Perhaps some extensions to `lib/github/github/server_mock.ex` would help?  (This also affects my changes in #392 fwiw...)

### DIFF
--- a/lib/github/github/server_mock.ex
+++ b/lib/github/github/server_mock.ex
@@ -29,6 +29,7 @@ defmodule BorsNG.GitHub.ServerMock do
       ...>   ] },
       ...>   {{:installation, 91}, 14} => %{
       ...>     branches: %{},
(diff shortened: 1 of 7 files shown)


couchand:
After spending a few minutes thinking about it, I decided it wouldn't be that hard to just go ahead and update the server mock to save the synthesized commits.  Let me know what you think.

Issue #374, opened 2018-03-21
Inconsistent history ordering


JensRantil:
See attached screenshot:

<img width="948" alt="screen shot 2018-03-21 at 21 28 12" src="https://user-images.githubusercontent.com/730342/37735611-dd547d90-2d4e-11e8-9f81-080e8900d8b1.png">

Expected is that it's ordered by time first.

PR #394, opened 2018-04-06
Sort log by time as well as date

couchand:
Previously, the log was sorted by date, but not by time
within a given date.  This corrects that issue.

Fixes: #374


notriddle:
timeout

bors r+

Issue #381, opened 2018-03-26
set custom committer e-mail for merge commits


couchand:
It would be nice to be able to configure an e-mail address to use for the merge commits.  This would support [this workaround](https://platform.github.community/t/commit-avatars-for-bots/2332/6) to make the commits show up with the bot's avatar rather than <img width="24" alt="screen shot 2018-03-26 at 5 37 33 pm" src="https://user-images.githubusercontent.com/793969/37934507-68b61bf2-311c-11e8-81b6-2863ffb6c15b.png">.

Issue #343, opened 2018-01-30
Push commits as a user other than bors-ng


Phrohdoh:
I'd like commits to master to belong to my user (@Phrohdoh).

Is this currently possible? If not is it planned?

notriddle:
Bors-NG doesn't support that, yet. It could be added pretty easily, if you don't mind GitHub saying that "Phrohdoh authored and bors committed" (having it hide bors entirely would be more complicated, since it would have to log in with your user instead of the installation token). It hasn't been on the planning backlog, because you're the first person who asked for it.

Not sure if you're worried about this, but https://github.com/bors-ng/bors-ng/graphs/contributors doesn't misattribute all the work to bors or anything. GitHub ignores merge commits when it generates the contributor boards.

jeffparsons:
I'm currently facing an issue related to this. I'm trying to get a private instance of Bors working with the [Buildkite CI system](https://buildkite.com), which in many cases requires that the email address associated with a commit is a "verified email" of one of the users in your Buildkite organisation.

I've opened an issue in the Buildkite organisation for discussion: https://github.com/buildkite/feedback/issues/313

But I'm really not sure what the best place is for a fix/workaround. I could imagine any of these:

- **GitHub:** let you associate a real email address with a GitHub application
- **Bors:** allow committing as another user
- **Buildkite:** some general way to play nice with GitHub applications

The other option I've considered is trying to set Bors up as an "OAuth App", because then it'd be a "real" GitHub user making the commits, and the problem would just go away.

(Here's the options:)

![screen shot 2018-02-12 at 1 05 21 pm](https://user-images.githubusercontent.com/321861/36081631-7578155e-0ff5-11e8-9f3c-e8fba3477e29.png)

Does anyone know if I'd be totally wasting my time trying that approach? If I had to make some changes to Bors to make it work, would you want to have them upstream? (I'd probably take a very different approach if this isn't a configuration you're interested in having in your tree.)

notriddle:
Using the git data API, bors-ng can create commits that are associated with about email address that it wants (though GitHub will label them as "X authored and Y committed"). I'd be perfectly happy to accept changes that make it so that.

I'd rather not not change it to an oauth app, because there are a bunch of APIs differences between oauth apps and github apps.

PR #396, opened 2018-04-06
Enable committer name & e-mail configuration

couchand:
This change adds a table to the bors.toml file to configure
the committer details for Bors merge commits. By adding a
committer table with name and email keys, the commits
created on GitHub will show up as authored by that user.

Fixes: #343
Fixes: #381 

Example:

    [committer]
    name = "Andrew Couch"
    email = "couchand@noreply.github.com"

### DIFF
--- a/lib/worker/batcher/bors_toml.ex
+++ b/lib/worker/batcher/bors_toml.ex
@@ -15,7 +15,12 @@ defmodule BorsNG.Worker.Batcher.BorsToml do
     timeout_sec: (60 * 60),
     required_approvals: 0,
     cut_body_after: nil,
-    delete_merged_branches: false
+    delete_merged_branches: false,
+    committer: nil
+
+  @type tcommitter :: %{
+    name: binary,
+    email: binary}
 
   @type t :: %BorsNG.Worker.Batcher.BorsToml{
     status: [binary],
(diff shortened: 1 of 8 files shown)


notriddle:
I hate to be a nitpicker, but could this be made a repository-level setting instead of (or in addition to) an instance-wide setting? Doing it in env variables like this prevents anybody from changing the committed-as email and name while still using the public instance.

notriddle:
Yeah

couchand:
Well that was a bit more complicated than I expected, but I think this should do it.

lib/worker/batcher/bors_toml.ex
```diff
@@ -24,24 +29,42 @@ defmodule BorsNG.Worker.Batcher.BorsToml do
...
     :cut_body_after |
+    :committer_details |
     :empty_config |
     :parse_failed
 
+  defp to_map(toml) do
+    toml
+      |> Enum.map(fn {key, val} -> {String.replace(key, "-", "_"), val} end)
+      |> Map.new()
+  end
```
notriddle: Good call factoring this out into its own function :+1:

notriddle:
I had to update the pull request body to actually say what the thing now does:

Old version:

> This change creates two environment variables to configure
the committer details for Bors merge commits.  By setting
COMMITTER_NAME and COMMITTER_EMAIL, the commits created in
GitHub will show up as authored by that user.

New version:

> This change adds a table to the bors.toml file to configure
the committer details for Bors merge commits.  By adding a
`committer` table with `name` and `email` keys, the commits
created on GitHub will show up as authored by that user.

couchand:
Just knocking out all the easy issues.  Oh neat, I see you already deployed since the latest merge.

notriddle:
You mind updating the docs page, too, since you're adding new config options? https://bors.tech/documentation/

PR #410, opened 2018-04-24
Allow mere members to see the actual project pages

### LINKED ISSUE #408: Listed repository yields a 404
I just tested out #407. Good stuff! 🙏  By adding the user as a member the repository is now

 * included in dashboard.
 * listed under `/repositories`.

That said, the repository (to which I've added the member) under `/repositories` links to a 404. My proposal is that at least `/repositories/3` (or whatever identifier the repo has) is visible and possibly history and settings are presented with read-only permissions.

What do you think of this?

notriddle:
Fixes #408

### DIFF
--- a/lib/database/context/permission.ex
+++ b/lib/database/context/permission.ex
@@ -37,6 +37,19 @@ defmodule BorsNG.Database.Context.Permission do
     true
   end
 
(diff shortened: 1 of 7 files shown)


notriddle:
Bors retry

PR #415, opened 2018-04-27
Allow an installation to sync repos directly

### LINKED ISSUE #187: Deal with missed webhooks sensibly
Bors-NG relies extensively on GitHub webhooks to work. Unfortunately, if GitHub isn't able to deliver a webhook (which could be their fault or our fault), they don't retry later.

# Webhooks that GitHub sends to bors; checked boxes have a fault recovery plan already

- [ ] **integration_installation** / **integration_installation_repositories**: These webhooks are sent from GitHub to Bors-NG when repositories are added or removed.

    In the case where a user is adding a repo, they will probably try to use bors immediately afterward (either by opening the dashboard or sending a command). Polling would not be sufficient in this case, because we don't want to poll more than once every hour or so, but when bors receives a request or a webhook related to a repo that it doesn't know about, it can pull down the info on-the-fly. (The hard part is doing this without potentially becoming a traffic multiplier for script kiddies with an agenda against GitHub).

    In the case where a user is removing a repo, we can't determine that something went wrong without polling. Since there's usually not a pressing need to remove a repo from our database (GitHub will already have stopped delivering any other webhooks, and there probably aren't any running builds when a user removes bors), we can use a daily polling job to pick up dropped repos. Daily polling can also pick up on dropped add-repo notifications, if a user adds a repo without immediately trying to use it.

- [X] **pull_request**: We have several ways to recover from dropped PR updates. There's a "Synchronize" button exposed to users, and when we get a notification about a PR that doesn't exist, we pull down its info to fill out the gap. The biggest problem with this is the "synchronize" action, which could cause a reviewer to accidentally r+ a stale version of the patch (because of the lack of any explicit ordering between "synchronize" and "issue_comment," the only correct way to fix this bug is to allow reviewers to name the commit that they want to approve).

- [X] **status**: Dropped status notifications are handled by polling at the Batcher level.

- [X] **pull_request_review** / **pull_request_review_comment** / **issue_comment**: Bors punts this to the user, by acknowledging their comment with either another comment or a Status API notification and giving them the opportunity to retry. It does not poll, and, having thought about it, I have no plans to add comment polling:
    * The Status API acknowledgement is necessary, anyway, so that users can detect if bors is completely down.
    * Scraping all the pull requests for new comments is an easy way to blow our rate limit, and if we don't scrape them frequently enough, the users will retry dropped commands anyway (because they won't get the status acknowledgement before the own personal timeout is exceeded).

# Other webhooks

- [ ] **GitHub notifies CI system of branch changes**: Travis CI and AppVeyor use the same at-most-once delivery as everybody else, and they don't seem to do any polling more than once a day. If their build never starts, bors will eventually time out (failing the build, but perhaps retrying). This can be fixed with something like a five-minute timeout for all "Build Started" status notifications, so that bors will re-do the merge commit to make GitHub try again, or we can work around it by using actual CI-specific API access.

- [ ] **CI system notifies GitHub of build results**: I have no idea if Travis or AppVeyor have any kind of retry behavior for their status notifications. However, if they don't, it's the same as the previous bullet; either we set up direct API access to the CI provider, we retry the entire build, or we do nothing and let it time out.

notriddle:
Fixes #187

(fixing the "other" webhooks is a separate issue)


notriddle:
I should add an admin interface to force a sync before merging this, thinking about it.

Issue #364, opened 2018-03-12
Bors says ":-1: Rejected by code reviews" even though the PR was approved


heinrich5991:
See https://github.com/ddnet/ddnet/pull/1079#issuecomment-372496230. Probably caused by #362.

grahamc:
Ack, I'm sorry! Looks like we've got a PR from notriddle to patch it up. LGTM, should be merged shortly.

heinrich5991:
Thanks, works.

vlad20012:
I've got the same issue at https://github.com/intellij-rust/intellij-rust/pull/2405

JensRantil:
We are seeing the same issue in a private repository:

![op_update_by_mathewclegg_ _pull_request__8562_ _tink-ab_tink-backend23](https://user-images.githubusercontent.com/730342/40412756-76cebd90-5e74-11e8-9f7a-2d971a0c8eaf.png)

![op_update_by_mathewclegg_ _pull_request__8562_ _tink-ab_tink-backend2](https://user-images.githubusercontent.com/730342/40412759-7b1fe194-5e74-11e8-9192-be6d83435a71.png)

Let us know if you need any additional feedback.

JensRantil:
Also, I propose this issue gets reopoened.

JensRantil:
This is still an issue for us and we keep experiencing it a couple of times per day. I've scrolled back a couple of hours back in our webhook history and it doesn't look like we've had any webhook exceptions related to the pull request. So far, I the workaround is to close the pull request and reopen the pull request, have the second person approve it and then `bors r+`. (cc @qinjin)

JensRantil:
I should also mention that this bug also happens if there has only been a single (approval) review. That is, it's not related to multiple "comment reviews" and a single "approve" review.

Let me know if I can give more information to determine what the issue is.

notriddle:
@JensRantil 

Pull request #429 should help with your problem. Can you please update to the newest bors-ng, and let me know if you're still able to reproduce this bug?

JensRantil:
Sweet! 🙏 We haven't had a way to reproduce the bug strictly, but I will share with my team that the bug hopefully is fixed and ask them to get back to me if they experience it again.

notriddle:
I'll close it unless anybody else has run into this problem (#435 should've fixed it).

JensRantil:
[...]

PR #435, opened 2018-06-28
Ignore commented reviews

notriddle:
This seems to be the problem that showed up in
https://github.com/meowpub/meow/pull/11

CC #364


notriddle:
Bors r+

PR #442, opened 2018-08-05
Fix the git clone part of the dev Dockerfile

notriddle:


notriddle:
n.b. to make sure I don't break people who were using 4000, like the old scripts exposed, the new scripts bind it to both ports.

bors r+

PR #447, opened 2018-08-15
Optimize the SVG icon

notriddle:
https://jakearchibald.github.io/svgomg/

Also made the icon square.


notriddle:
@JensRantil That's cool, but now that I come back here, @ntim found a way to squeeze it even more (there were a bunch of font-related styles in there for no reason).

notriddle:
bors r= @JensRantil

PR #448, opened 2018-08-17
Do not authenticate the wobserver API

notriddle:
Wobserver uses its own authentication key, so we can forward anything in and let it take care of it.

Fixes https://forum.bors.tech/t/accessing-wobserver/201/8


unicolet:
@notriddle thanks!

PR #451, opened 2018-08-17
Further raise the PR timeout

notriddle:


notriddle:
Julia added their entire organization to bors-ng. This caused the background syncer to time out trying to pull down a list of all active PRs and repositories.

I raised the timeout, which allowed us to suck down all that stuff right now and get things working, but what we really need is smarter GitHub API rate limiting, where each installation gets its own GitHub.Server process.

bors r+

PR #504, opened 2018-10-02
Keep GitHub App configuration up-to-date

michaelbaudino:
This PR updates the README instructions for GitHub App webhooks
configuration to match GitHub web UI regarding a checkbox that seems to
have changed: [`Check run`](https://developer.github.com/v3/activity/events/types/#checkrunevent) and [`Check suite`](https://developer.github.com/v3/activity/events/types/#checksuiteevent) are now 2 distinct checkboxes.

screenshot_2018-10-02 build software better together

PS: I saw that the `check_suite` webhook is [silently ignored at the moment](https://github.com/bors-ng/bors-ng/blob/638d76597872829b6af276873db7a6e1143b73c2/lib/web/controllers/webhook_controller.ex#L197-L199), but since we do expect to receive it, I thought it made sense to configure it. Feel free to let me know if it should actually not been configured at all.

Thanks for this awesome tool, everybody :heart:

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,8 @@ For each of these sections, set the following overall section permissions and ch
   - *Organization* ( User invited to, added to, or removed from an organization)
 - *Organization projects*: No access
 - *Checks*: Read & Write
-  - *Check run* (CheckSuite created from the API)
+  - *Check run* (Check run created from the API)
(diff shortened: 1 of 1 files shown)


notriddle:
bors r+

Good catch. Thanks!

Issue #434, opened 2018-06-20
bors retry race condition


hackaugusto:
`bors retry` will update the staging branch even if there is a build currently running. If a job has not started yet the expected commit will be deleted and the build will fail.

[job started before retry](https://travis-ci.org/raiden-network/raiden/jobs/394588783)
[`bors retry`](https://github.com/raiden-network/raiden/pull/1583#issuecomment-398778838)
[job started after retry](https://travis-ci.org/raiden-network/raiden/jobs/394588784)

andschwa:
This appears to happen with lots of `bors try` in a row too.

It seems that as soon as Bors sees that a `bors try` build has started (but not finished), it thinks it's free to force-push the `trying` branch. However, our build system (Jenkins / Blue Ocean) will start the job as soon as it can, though sub-jobs may still be waiting for executors to become available. That is, the `trying` branch is used by Jenkins to checkout the repo throughout the job's life, and really isn't free to be force-pushed until the job is completed.

This seems related, but maybe it warrants a new issue.

andschwa:
Our CI system is weighed down to the point that we had to stop using `bors try` since this happened every time, so we're stuck using the sandbox when testing CI changes. I do hope this gets fixed (though I'm not sure what the fix is as the issue truly lies in Jenkins).

pietrocaselani:
We are facing the same problem using `bors try`. Any updates on this? For now, I think we will stop using `bors try`

notriddle:
The commenters in here are talking about two issues.

One of them, the `bors try` race issue, is now fixed.

The other one, where bors r+ followed by a cancel followed by another job (either another r+ or just the next queued item running) caused Travis to start throwing up errors, is not a bug. It's caused by the fact that bors doesn't wait for canceled builds to be finished at the provider end before starting the next one. Since it's canceled at bors' end, I don't see why it matters that it's not able to finish. Won't fix.

PR #582, opened 2019-01-25
Queue up attempts rather than running them all at once

notriddle:
https://forum.bors.tech/t/bors-try-doesnt-wait-untill-last-build-finishes/272

https://github.com/bors-ng/bors-ng/issues/434#issuecomment-434856549

### DIFF
--- a/icon/bors-path.svg
+++ b/icon/bors-path.svg
@@ -14,8 +14,11 @@
    viewBox="0 0 47.825287 13.123333"
    version="1.1"
    id="svg8"
-   inkscape:version="0.92pre4 15278"
-   sodipodi:docname="bors-path.svg">
+   inkscape:version="0.92.3 (2405546, 2018-03-11)"
+   sodipodi:docname="bors-path.svg"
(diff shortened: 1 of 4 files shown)


andschwa:
Thank you so much @notriddle!

(Also you asked on the forum why people might be using `bors try` instead of CI auto-building PRs. In my case, we have two reasons: (1) we don't want to auto-build PRs from just anyone, so `bors try` gives us a bit of an authentication layer, and (2) if we built every PR whenever they're opened/updated our CI would be so bogged down it would be useless.)

notriddle:
(Please give your response on the forum, also, so that other people can see it.)

notriddle:
No problem.

(I also copied your reply over to the forum, so that other participants would be able to see it.)

andschwa:
Thank you! I don't have account there yet 😄

andschwa:
login

new

🤷‍♂️

notriddle:
Your adblocker has an outdated, or otherwise weird, filter policy.

andschwa:
Ah, that is true! Good to know. Thanks again 😄

Issue #568, opened 2019-01-03
Show a different error when there's already a `try` build running


notriddle:
Right now, bors writes this comment when a user runs `bors try` twice in a row.

> ### try
>
> Not awaiting review

That's a nonsense error message. It should probably say something more like this:

> ### try
>
> Already running

notriddle:
Here's where the message gets generated:

https://github.com/bors-ng/bors-ng/blob/4d609b3/lib/worker/attemptor.ex#L81-L84

PR #590, opened 2019-02-03
Updating message when review is already running

mimimalizam:
This PR replaces message `Not awaiting review` with `Already running a review` for Atemptor and Batcher messages when another review is in progress. It should resolve https://github.com/bors-ng/bors-ng/issues/568. Please let me know if there is anything else left to be changed. Also if suggested Batcher change is excess, I can remove it :) .


mimimalizam:
tnx @notriddle!

Issue #561, opened 2018-12-11
Crash batch push 422


ForsakenHarmony:
Could this happen because master is protected? 

```
{{:badmatch, {:error, :push, 422}},  [    {BorsNG.Worker.Batcher, :complete_batch, 3,     [file: 'lib/worker/batcher.ex', line: 386]},    {BorsNG.Worker.Batcher, :maybe_complete_batch, 1,     [file: 'lib/worker/batcher.ex', line: 373]},    {BorsNG.Worker.Batcher, :handle_cast, 2,     [file: 'lib/worker/batcher.ex', line: 81]},    {:gen_server, :try_dispatch, 4,     [file: 'gen_server.erl', line: 637]},    {:gen_server, :handle_msg, 6,     [file: 'gen_server.erl', line: 711]},    {:proc_lib, :init_p_do_apply, 3,     [file: 'proc_lib.erl', line: 249]}  ]}
```

zegl:
This also seems to happen if is an open negative review ("request changes") open on the PR. 

It would be nice if bors was more helpful in this situation, and posted a message as a PR comment, instead of crashing. Users that doesn't have bors reviewer access doesn't have access to look at the crash logs either, which makes this error frustrating for the user.

ForsakenHarmony:
Well I figured out it happens when reviews are required and you try to merge with bors before approving

umamaistempo:
I'll be taking a look on this issue. Hopefully I'll be shipping a fix on friday or monday (tomorrow is a holiday here)

umamaistempo:
This kind of problem should not occur anymore. In case anything similar or derived from this error occurs, feel free to create a new issue and tag me

umamaistempo:
@notriddle please close this

notriddle:
Sure, okay.

If anyone else runs into it, @ me so I can reopen.

Issue #587, opened 2019-01-30
Add option to disable "rejected by code reviews" behavior


matklad:
Hi!

It seems that bors does not accept r+ if there's a "rejected" PR review: https://github.com/rust-analyzer/rust-analyzer/pull/705#issuecomment-459064754.

Is there an option to disable this behavior? I haven't fond one in bors.toml. I want to disable it because I don't personally use approval/disapproval mechanism, but some contributor might give a negative review which I'll have to override (not because I disagree with review, but because concerns were addressed :) )

Additionally, perhaps it makes sense to make this opt-in, rather than on by default?

PR #656, opened 2019-06-05
Make merge fail explicitly when an approved PR also has changes requested

umamaistempo:
Currently here at Tink AB, frequently the following situation happens:

PR
- Person request changes
- New commit fix changes
- Gets approval and ask bors to merge
- Bors never says anything and nothing seems to happen
- Someone notices this and dismiss stale "changes requested"
- Ask for bors to merge and it works wonders

This is also correlated to #561 

This PR aims to solve this problem by getting bors to explicitly refuse (providing message) when a merge is requested but there are changes requested.

Some minor types were added (as I was using them to explore the code to find out where the problem was). They can be merged as I intend to provide more typespecs in following PRs

Edit:
Assertions to test:
- [x] has approvals, has disapprovals, should fail and return msg
- [ ] ~has approvals, has disapprovals, disapproval ignored in toml, should succeed~
- [ ] ~no approval, has disapprovals, disapproval ignored in toml, should succeed~

Edit2:
There's no semantics for "ignore disapprovals", the only configs seems to be either require approvals (at which time disapprovals are considered) or doesn't (where disapprovals are also ignored).

This semantics seems strange TBH. I'll keep them working but I believe that the possible branchs should be

- Required Approvals(int, 0 by default)
- Ignore disapprovals(bool, false by default)

So a 0 approval, 1 disapproval when approvals are not required does not seem to make sense in the usual case

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -383,7 +383,7 @@ defmodule BorsNG.Worker.Batcher do
     end
   end
 
-  @spec complete_batch(BorsNG.Database.Status.state, BorsNG.Database.Batch.t, term) :: term
+  @spec complete_batch(Status.state, Batch.t, [Status.t]) :: :ok
   defp complete_batch(:ok, batch, statuses) do
     project = batch.project
     repo_conn = get_repo_conn(project)
@@ -539,13 +539,9 @@ defmodule BorsNG.Worker.Batcher do
     |> Enum.map(fn {context, _} -> context end)
     |> MapSet.new()
     |> MapSet.disjoint?(MapSet.new(toml.pr_status))
-    passed_review = if is_nil(toml.required_approvals) do
-      :sufficient
-    else
-      repo_conn
-      |> GitHub.get_reviews!(patch.pr_xref)
-      |> are_reviews_passing(toml.required_approvals)
-    end
+    passed_review = repo_conn
+    |> GitHub.get_reviews!(patch.pr_xref)
+    |> are_reviews_passing(toml)
     case {passed_label, passed_status, passed_review} do
       {true, true, :sufficient} -> :ok
       {false, _, _}             -> {:error, :blocked_labels}
@@ -555,13 +551,22 @@ defmodule BorsNG.Worker.Batcher do
     end
   end
 
-  defp are_reviews_passing(reviews, required) do
-    %{"CHANGES_REQUESTED" => failed, "APPROVED" => passed} = reviews
-
-    case {failed, passed} do
-      {failed, _} when failed > 0 -> :failed
-      {_, approved} when approved >= required -> :sufficient
-      {0, _} -> :insufficient
+  @spec are_reviews_passing(map, Batcher.BorsToml.t) :: :sufficient | :failed | :insufficient
+  defp are_reviews_passing(reviews, toml) do
+    %{"CHANGES_REQUESTED" => failed, "APPROVED" => approvals} = reviews
+    required? = is_integer(toml.required_approvals)
+    approvals_needed = required? && toml.required_approvals || 0
+    approved? = approvals >= approvals_needed
+    failed? = failed > 0
+    cond do
+      not required? ->
+        :sufficient
+      failed? ->
+        :failed
+      approved? ->
+        :sufficient
+      required? ->
+        :insufficient
     end
   end
 
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -383,6 +383,7 @@ defmodule BorsNG.Worker.Batcher do
     end
   end
 
+  @spec complete_batch(BorsNG.Database.Status.state, BorsNG.Database.Batch.t, term) :: term
   defp complete_batch(:ok, batch, statuses) do
     project = batch.project
     repo_conn = get_repo_conn(project)
@@ -558,7 +559,7 @@ defmodule BorsNG.Worker.Batcher do
     %{"CHANGES_REQUESTED" => failed, "APPROVED" => passed} = reviews
 
     case {failed, passed} do
-      {failed, 0} when failed > 0 -> :failed
+      {failed, _} when failed > 0 -> :failed
       {_, approved} when approved >= required -> :sufficient
       {0, _} -> :insufficient
     end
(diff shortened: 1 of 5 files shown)


umamaistempo:
Considering returning different message for this case. Something along "Branch can't be merged because it has an open request for changes". What do you think ?

lib/worker/batcher.ex
```diff
@@ -538,13 +539,9 @@ defmodule BorsNG.Worker.Batcher do
...
-    passed_review = if is_nil(toml.required_approvals) do
-      :sufficient
-    else
-      repo_conn
-      |> GitHub.get_reviews!(patch.pr_xref)
-      |> are_reviews_passing(toml.required_approvals)
-    end
+    passed_review = repo_conn
+    |> GitHub.get_reviews!(patch.pr_xref)
+    |> are_reviews_passing(toml)
```
umamaistempo: Note: I wrote this way to keep with the style, but personally I think that it's more readable when we indent it in the next block:

```
review_status = 
  repo_conn
  |> GitHub.get_reviews!(patch.pr_xref)
  |> are_reviews_passing(toml)
```

lib/worker/batcher/state.ex
```diff
@@ -5,23 +5,23 @@ defmodule BorsNG.Worker.Batcher.State do
   and emits a batch state.
   """
 
-  @typep n :: BorsNG.Database.Status.state_n
-  @typep t :: BorsNG.Database.Status.state
+  @typep status :: BorsNG.Database.Status.t
```
umamaistempo: this one was incorrect. `state_n` is how the `state` was being stored in the DB. The thing that the function receives is actually a list of `%Status{}` as can be seen

lib/worker/batcher.ex
```diff
@@ -554,13 +551,22 @@ defmodule BorsNG.Worker.Batcher do
...
+  @spec are_reviews_passing(map, Batcher.BorsToml.t) :: :sufficient | :failed | :insufficient
+  defp are_reviews_passing(reviews, toml) do
+    %{"CHANGES_REQUESTED" => failed, "APPROVED" => approvals} = reviews
+    required? = is_integer(toml.required_approvals)
+    approvals_needed = required? && toml.required_approvals || 0
+    approved? = approvals >= approvals_needed
+    failed? = failed > 0
+    cond do
+      not required? ->
+        :sufficient
```
umamaistempo: I _really_ want to remove this. I included this clause to keep backwards compatibility (if `required_approvals` is not present, we don't check for failure) but I personally find this strange

lib/worker/batcher.ex
```diff
@@ -554,13 +551,22 @@ defmodule BorsNG.Worker.Batcher do
...
-    %{"CHANGES_REQUESTED" => failed, "APPROVED" => passed} = reviews
-
-    case {failed, passed} do
-      {failed, 0} when failed > 0 -> :failed
-      {_, approved} when approved >= required -> :sufficient
-      {0, _} -> :insufficient
+  @spec are_reviews_passing(map, Batcher.BorsToml.t) :: :sufficient | :failed | :insufficient
+  defp are_reviews_passing(reviews, toml) do
+    %{"CHANGES_REQUESTED" => failed, "APPROVED" => approvals} = reviews
+    required? = is_integer(toml.required_approvals)
```
umamaistempo: considering renaming to `review_required?`

lib/worker/batcher.ex
```diff
@@ -554,13 +551,22 @@ defmodule BorsNG.Worker.Batcher do
...
 
-  defp are_reviews_passing(reviews, required) do
-    %{"CHANGES_REQUESTED" => failed, "APPROVED" => passed} = reviews
-
-    case {failed, passed} do
-      {failed, 0} when failed > 0 -> :failed
-      {_, approved} when approved >= required -> :sufficient
-      {0, _} -> :insufficient
+  @spec are_reviews_passing(map, Batcher.BorsToml.t) :: :sufficient | :failed | :insufficient
+  defp are_reviews_passing(reviews, toml) do
```
umamaistempo: considering renaming to `reviews_status` as the current function name sounds like something that returns a boolean

lib/worker/batcher.ex
```diff
@@ -383,6 +383,7 @@ defmodule BorsNG.Worker.Batcher do
     end
   end
 
+  @spec complete_batch(Status.state, Batch.t, [Status.t]) :: :ok
```
umamaistempo: I did not explore the code enought to find out if by design it is impossible for `waiting` status batch to hit the `maybe_complete_batch/1` fun; if there's no safety net there, then it's possible for this function to receive such state and that would raise a function argument error (we're only treating the cases of `ok`, `error` and `running`

lib/worker/batcher.ex
notriddle: Sure, sounds good to me.

lib/worker/batcher.ex
notriddle: Agree.

lib/worker/batcher.ex
notriddle: [Other users asked for it](https://github.com/bors-ng/bors-ng/issues/587). @matklad

I think, at least in the case under discussion, that the user who requested changes was a non-collaborator. If you can make sure that only collaborator's reviews are counted, then it could probably be removed.

Or we could always add a `bors r+ force` option, to more or less mimic GitHub's "turn the merge button red" behavior. But that should probably go through an RFC.

ForsakenHarmony:
does this also work if the master branch is protected, but there's no reviews at all?

lib/worker/batcher.ex
matklad: Yeah, I would prefer that there be an option to treat bors commands as a single source for whether the PR should be merged. At least in small-scale open-source projects, dismissing requests for changes seems like a busy work. It seems quite unlikely the r+er *accidentally* misses request for changes, and it is quite probably that they want to deliberately ignore it without extra clicks.

lib/worker/batcher.ex
umamaistempo: So, I will merge this PR now, but then, to improve documentation, I'll make the type

```
@type staged_state :: :ok | :error | :running
```

So we can specify it here, what do you think ? :^)

Issue #639, opened 2019-04-29
Repository listing order improvement


JensRantil:
I have a repository with a long list of pull requests. Every time I visit https://www.mybors.com/repositories/42 it lists >40 pull requests and my most common use-case is to look at the "Running" and/or "Waiting to run". Since those lists generally are bound to be shorter, how about moving the _before_ "Awaiting review". My proposal is the following order:

 1. Running
 2. Waiting to run
 3. Awaiting review

JensRantil:
Example of how it looks today for me:

![image](https://user-images.githubusercontent.com/730342/57465234-7bcd2b00-727e-11e9-9422-7d38a12db496.png)

umamaistempo:
Forgot to incluse a `closes #639` on the PR. The #658 PR "fixes" this by prioritizing transient states on presentation (waiting and running). @JensRantil this can be closed

JensRantil:
Sweet! Thanks! 😍

PR #658, opened 2019-06-07
Reorder presentation of batches

umamaistempo:
Based on #639 

Rationale: For someone administering a bors instance, usually it should be more important to see the build status instead of what still needs to be reviewed (as this would usually be done by checking the github notifications). And considering `:running | :waiting` is a short transitive state, the list of such batches is way shorter and will disappear quickly, so not affecting the use case of people that uses the web view to see what needs reviewing.

Note: also added an order to other queries on the batch model as to ensure consistency (ie: without an explicit ordering on postgres - and other dbs -, there's no order guarantee, so results are somewhat non-deterministic)

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -655,17 +655,22 @@ defmodule BorsNG.Worker.Batcher do
   end
 
   defp put_incomplete_on_hold(repo_conn, batch) do
-    batches_query = batch.project_id
+    batches = batch.project_id
     |> Batch.all_for_project(:running)
     |> where([b], b.id != ^batch.id and b.priority < ^batch.priority)
-    Status
-    |> join(:inner, [s], b in ^batches_query, s.batch_id == b.id)
-    |> Repo.delete_all()
-    batches_query
     |> Repo.all()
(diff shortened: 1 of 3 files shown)


lib/worker/batcher.ex
```diff
@@ -655,17 +655,22 @@ defmodule BorsNG.Worker.Batcher do
...
-    |> Repo.all()
-    |> Enum.each(&send_status(repo_conn, &1, :delayed))
-    Repo.update_all(batches_query,
-      set: [state: :waiting])
+
+    
+    Enum.each(batches, &send_status(repo_conn, &1, :delayed))
+    Batch
+    |> where([b], b.id in ^ids)
+    |> Repo.update_all(set: [state: :waiting])
```
umamaistempo: I practice, the changes to this function, if affecting DB performance, will improve it, because we were going to fetch the batches before, and by using the `ids` directly instead of the more complex query for join might benefit the query planner (unless it's properly caching and identifying that the resultset would be the same, in such case there's virtually no difference between both).

So, this change does not degrade performance nor readability :>

notriddle: Agreed.

PR #659, opened 2019-06-07
Do not repeat projects on Dashboard when simultaneously Member and Reviewer

### LINKED ISSUE #657: When a user is both a Reviewer and a Member, the repository is listed twice in their repositories list
To some extent, it doesn't make sense to have the same user be a Reviewer and a Member, but it's not clear that it's harmful.

Some mechanism to avoid listing the repo twice (by some solution that just deals with a user being both a member and reviewer, or by some mechanism which prevents/migrates/etc users that are in both to just be reviewers) would be nice to avoid confusion.

umamaistempo:
Closes #657


umamaistempo:
Ops, thought we were PgSQL exclusively, but it seems we support MySQL. Will revert last commit then

PR #660, opened 2019-06-07
Encode github URIs to avoid bad request

### LINKED ISSUE #626: Adding a username with a leading space to the dashboard causes internal server error
I've tried to add `username` to reviewers, but I accidently typed ` username`, and that resulted in an internal server error. This was for rust-analyzer repo just now, there might be some useful logs.

umamaistempo:
Closes #626

### DIFF
--- a/lib/github/github.ex
+++ b/lib/github/github.ex
@@ -183,7 +183,7 @@ defmodule BorsNG.GitHub do
   def get_user_by_login!(token, login) do
     {:ok, user} = GenServer.call(
       BorsNG.GitHub,
-      {:get_user_by_login, token, {login}},
+      {:get_user_by_login, token, {String.trim(login)}},
       Confex.fetch_env!(:bors, :api_github_timeout))
     user
   end
--- a/lib/github/github/server.ex
+++ b/lib/github/github/server.ex
@@ -327,7 +327,7 @@ defmodule BorsNG.GitHub.Server do
   ) do
     "token #{token}"
     |> tesla_client()
-    |> Tesla.get!("/users/#{login}")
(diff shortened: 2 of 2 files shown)


lib/github/github.ex
```diff
@@ -183,7 +183,7 @@ defmodule BorsNG.GitHub do
   def get_user_by_login!(token, login) do
     {:ok, user} = GenServer.call(
       BorsNG.GitHub,
-      {:get_user_by_login, token, {login}},
+      {:get_user_by_login, token, {String.trim(login)}},
```
umamaistempo: I only trimmed in this place because I am sure GH does not allow such so I will consider it a typo. 

On the other hand, I'm not doing any assumption on any other user input

notriddle: Just to make sure, when you do this, you've made sure that it can't result in us accidentally creating a record in the `users` table with a leading space?

umamaistempo: Good catch, will confirm it and add a maybe-redundant trimmer to the users model

umamaistempo: On all cases that we use this function, we use it's return as the username when inserting the record. This makes this problem something that _should_ not happen (unless we change the code somehow without noticing it) and also by incident handles the potential bug with casing (`"NOTriddle" =/= "notriddle"`)

umamaistempo: So we can either:
a) leave it with the trim to handle the case where the user copy-paste with an space
b) remove this trim and put an additional line on the error saying that it might have occurred by the present of whitespace

Other than that, LGTM :)

notriddle: Nah, let's just go ahead and merge as-is.

PR #661, opened 2019-06-08
Add note about the renaming of integration ID in README

samueldple:


samueldple:
Sorry - didn't realise my editor had reformatted the rest of it too. Will resubmit.

PR #664, opened 2019-06-11
Sort repositories on Dashboard by name

### LINKED ISSUE #663: Feature request: Sort the list of repos alphabetically
I have installed Bors on quite a few repositories. Currently, when I go to the list of repositories ([https://app.bors.tech/repositories](https://app.bors.tech/repositories)), the repositories are sorted in (I believe) the order in which Bors was installed. This makes it a little difficult to navigate a long list of repositories.

Could we instead sort the list of repositories in descending alphabetical order?

umamaistempo:
Closes #663

Decided to consolidate it using a `group_by`, this reduces the amount of data returned from DB, the amount of handshakes and the cost of enumerating it, while, in the meantime, the DB can optimize the resultset :>

Might add later some test to check this expectation (ie: add 3 projects, set as reviewer and member of them - in one of them, both - and check that the resultset is the expected)

Edit: Note that on #659 the `order_by: p.repo_xref` was added to ensure that the repo list is consistent (there is no implicit guarantee on ordering for RDBMS without an explicit ordering request). Since we need an order, as mentioned on #663, sorting alphabetically is what would be most helpful for our users as it's easier to find what you expect

Edit2: Thinking about this, it would also probably be a good addition to order the ProjectController.show/4 results. I think that ordering it by `patch.pr_xref DESC` would be the most useful ordering. What do you think ?

### DIFF
--- a/lib/database/context/dashboard.ex
+++ b/lib/database/context/dashboard.ex
@@ -11,7 +11,7 @@ defmodule BorsNG.Database.Context.Dashboard do
     from(p in Project,
       join: l in LinkUserProject, on: p.id == l.project_id,
       where: l.user_id == ^user_id,
-      order_by: p.repo_xref)
+      order_by: [asc: p.name, asc: p.repo_xref])
     |> Repo.all()
   end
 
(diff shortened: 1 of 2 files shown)


notriddle:
> Thinking about this, it would also probably be a good addition to order the ProjectController.show/4 results. I think that ordering it by patch.pr_xref DESC would be the most useful ordering. What do you think ?

Yeah, that might be a good idea; you could put it in another PR.

PR #682, opened 2019-07-02
Fix missing domain name in badge code

notriddle:


DilumAluthge:
Thank you! I was just trying to figure out how to fix that - didn't want to hard code `app.bors.tech` into the URL, because then it would break for all the people using private instances of bors.

notriddle:
bors retry

https://www.traviscistatus.com/incidents/g9xm3l02dknv

PR #687, opened 2019-07-05
Switch from brunch to webpack

notriddle:
Brunch isn't really being maintained any more,
and some of its deps are giving me security warnings.


notriddle:
Wait, no, I just found a problem.

bors r-

notriddle:
That should solve that.

bors r+

PR #700, opened 2019-07-11
Return error body upon failed push to GitHub

swist:
We were experiencing strange 422 errors when bors was trying to merge
a successful speculative merge. It was failing because of misconfigured
branch protection. Returning the body would trigger the :badmatch
exception with the error message


DilumAluthge:
I just wanted to say thank you to @swist and @notriddle for this patch! It has been tremendously helpful to me in understanding the causes of 422 error.

Issue #194, opened 2017-05-25
Support squash and merge as well as fast-forward and merge


jniles:
It would be nice if the commits in a PR could be squashed with bors instead of simply fast fowarded.  This would allow projects to accept PRs even if the contributor had not followed commit guidelines correctly, without having to ask the contributor to rebase and squash correctly.  I imagine this implementation would have a "merge type" flag in `bors.toml` that might change the default behavior to squash-and-merge.

notriddle:
Is this a duplicate of #138? Or would you still like the merge commit at the end, just with the user's PR commits squished?

jniles:
I had see #138, but wasn't sure if it included squashing the individual PR's commit history.  If I understand correctly, if a PR has two commits on it is merged, it will look like this:

```sh
Commit #2 - PR Author
|
Commit #1 - PR Author
|
Original Master Head
```

Ideally, it would be nice to squash the commits, 
```sh
Merge #PR (Comment #1 and #2 Inside)
|
Original Master Head
```

I can see that this might not be the `bors-ng` way though, if PRs are to be batched.  It also would destroy the contributor's commit history, since @bors[bot] would always be the author.  The goal though, would be to remove the burden of correct commit message syntax from the contributor since their messages would be collapsed anyway.

notriddle:
> I had seen #138, but wasn't sure if it included squashing the individual PR's commit history. If I understand correctly, if a PR has two commits on it is merged, it will look like this

Yeah, that's how I imagined #138 working,

> Ideally, it would be nice to squash the commits,

I see what you mean, now. Yeah, they're not the same thing. They both constitute "messing with the history," but they're not the same.

---

What we end up with is this matrix:

|             | Squash  | Don't     |
|---------|----------|----------|
| Rebase | ?            | #138     |
| Don't    | #194     | Current |

Both involve messing with the history, so they have some implementation overlap, but you could rebase the commits without squashing them (which means you don't need a merge commit) and you could squash the commits without rebasing them (which means you still need a merge commit).

Do you think we'd need to implement the whole matrix, though?

jniles:
Since I'm just getting my feet wet with bors-ng, I'm hesitant to recommend a lot of development in this area.  My original motivation for this was to make sure the history is clean, but having a history of `Merge #3`, `Merge #2`, `Merge #1` is likely not informative.

I'll get a little more experience playing around with this before coming up with a solid proposal.  I do appreciate your time considering it though!

[...]

PR #704, opened 2019-07-16
Allow Bors to use squash merging

MarkSRobinson:
This PR adds the ability for Bors to use a squash merge instead of the current default merge.

The rough operation is:
* For each Patch in a Batch
 -> Generate a title from PR title & number
 -> Generate a commit message from the commit message of all PR commits
 -> Use the Github merge API w/ "squash" option enabled
 -> Each PR is merged in turn into master.

This does not change the default behaviour.  Users must add `use_squash_merge = true` to their bors.toml file to enable this functionality.

This change is to address https://github.com/bors-ng/bors-ng/issues/194


notriddle:
First of all, I'd like to apologize for screwing up. We had just started a new, more formal, process for feature proposals, but I never actually mentioned it in the README. I've gone ahead and added these notes to README and CONTRIBUTING.

So here's what you'll need to do:

* We need to fix your commits. You've got `m********@p****.com` set as your email address in `Git`, but you don't have that email verified with GitHub. Go to https://github.com/settings/emails and add it as an alternative email address.

  **Why**: It'll give you an avatar and a link to your profile next to the commits on this PR, and it'll give you your spot in the [contributors](https://github.com/bors-ng/bors-ng/graphs/contributors) list.

* Go to https://forum.bors.tech/c/development/draft-rfcs and open a new topic there. You can log in with your GitHub Account, or you can register separately. Fill out the template.

  **Why**: The RFC template basically includes all the how-to documentation, but before we merge it. We've had trouble in the past with features being implemented in bors-ng, but nobody actually wrote up how they were supposed to be used, hence "the docs have to be written before we merge any code" becomes the new rule.

* After the RFC template is filled out, we'll wait about four weeks to give other people a chance to look at it.

  **Why**: There are a bunch of people who've wanted squash-merging in bors, but they might have different ideas about exactly what they want. Most importantly, I'm not one of them, so I'm not sure if what you're doing is actually a good way to implement this or not.

It sounds like a lot, but it shouldn't take more than an hour of your time --- total --- to do the first two steps, and the last step is probably just going to be waiting and maybe answering a few questions to clarify your docs.

notriddle:
Also, dialyzer failed https://tools.taskcluster.net/groups/OcLOXFRdQre596OGaNfFvw/tasks/OcLOXFRdQre596OGaNfFvw/runs/0/logs/public%2Flogs%2Flive.log#L1160

notriddle:
And ... thanks for fixing both of those things!

notriddle:
Obsoleted by https://github.com/bors-ng/bors-ng/pull/718

PR #725, opened 2019-08-09
[WIP] Parse and use CODEOWNER

MarkSRobinson:
Implementation of RFC: https://forum.bors.tech/t/bors-support-for-codeowners/357


lib/github/github.ex
```diff
@@ -188,6 +201,25 @@ defmodule BorsNG.GitHub do
     user
   end
 
+  @spec get_team_by_name(tconn, bitstring, bitstring) ::
```
umamaistempo: Please, instead of bitstring use `String.t`

lib/github/github.ex
```diff
@@ -26,6 +26,19 @@ defmodule BorsNG.GitHub do
   @type tcollaborator :: %{user: tuser, perms: tuser_repo_perms}
   @type tcommitter :: %{name: bitstring, email: bitstring}
 
+  @spec get_pr_files!(tconn, integer | bitstring) :: [BorsNG.GitHub.File.t]
```
umamaistempo: If I am not mistaken, pr_xref is integer only

lib/github/github.ex
```diff
@@ -188,6 +201,25 @@ defmodule BorsNG.GitHub do
...
+  def get_team_by_name(repo_conn, org_name, team_name) do
+    GenServer.call(
+      BorsNG.GitHub,
+      {:get_team_by_name, repo_conn, {org_name, team_name}},
+      Confex.fetch_env!(:bors, :api_github_timeout))
+  end
+
+  @spec is_user_on_team!(tconn, bitstring, integer) ::
+          boolean
+  def is_user_on_team!(repo_conn, username, team_id) do
```
umamaistempo: I am confused by this. According to Elixir convention, ending with an `!` means that it should throw an error when unable to go through the happy path. If it's returning a bool, it should end with `?`. Also according to Elixir conventions, you don't need to use `is_` when the function ends with `?`, so I'd suggest renaming `user_on_team?` or better yet, `belongs_to_team?` as it would be very fluent based on the context

lib/github/github.ex
umamaistempo: Note: this everywhere that `bitstring` appears

lib/github/github/file.ex
```diff
@@ -0,0 +1,41 @@
+defmodule BorsNG.GitHub.File do
+  @moduledoc """
+  The structure of GitHub pull request file
+  """
+
+  @type tjson :: map
```
umamaistempo: :thinking: well, the `tsomething` type names exist on github.ex file but aren't very common TBH. I think that just saying map on `from_json/1` is good enough

lib/github/github/file.ex
```diff
@@ -0,0 +1,41 @@
...
+  Convert from Poison-decoded JSON to a Pr struct.
+  """
+  @spec from_json!(tjson) :: [t]
+  def from_json!(json) do
+    {:ok, files} = from_json(json)
+    files
+  end
+
+  @doc """
+  Convert from Poison-decoded JSON to a Pr struct.
```
umamaistempo: I would remove the `poison-decoded` part because we can anytime change it from Poison to something like https://github.com/michalmuskala/jason.

Also it's not returning a PR struct, but a github file

lib/github/github/file.ex
```diff
@@ -0,0 +1,41 @@
...
+  """
+  @spec from_json(tjson) :: {:ok, [t]} | {:error}
```
umamaistempo: you forgot the second element returned on `{:error}`, it should be something like `from_json(input) :: {ok, [t]} | {:error, input} when input: map`

lib/github/github/reviews.ex
```diff
@@ -8,16 +8,30 @@ defmodule BorsNG.GitHub.Reviews do
...
       end)
+
+    # Get the list of users who have approved this PR
+    approvers = reviews
+                |> Enum.flat_map(fn {_uid, state} ->
+      case state do
+        # Ignore dismissed reviews
+        "DISMISSED" -> []
+        # Pass just the username onward
+        id -> [_uid]
```
umamaistempo: if you are using the variable, if possible, do not prepend it with an underscore `_`, this in elixir terms represents a binding/variable that you will not be using, and this convention makes some analysis ignore the binding

lib/github/github/reviews.ex
```diff
@@ -8,16 +8,30 @@ defmodule BorsNG.GitHub.Reviews do
...
       %{"state" => "COMMENTED"}, acc ->
         acc
-      %{"user" => %{"id" => uid}, "state" => state}, acc ->
+      %{"user" => %{"login" => uid}, "state" => state}, acc ->
         Map.update(acc, uid, state, fn _ -> state end)
       end)
+
+    # Get the list of users who have approved this PR
+    approvers = reviews
+                |> Enum.flat_map(fn {_uid, state} ->
```
umamaistempo: If you are just getting one value per entry, then you don't need to flat-map, a filter -> map would be better:

```elixir
approved_by =
  reviews
  |> Enum.reject(fn {_, state} -> state == "DISMISSED" end)
  |> Enum.map(fn {username, _} -> username end)
  |> Enum.uniq()
```

umamaistempo: Also, considering that `reviews` is a map whose key is the uid, there's no way that an username/uid is repeated because a map imply in keys being unique, so the last operation is not necessary

lib/github/github/reviews.ex
```diff
@@ -8,16 +8,30 @@ defmodule BorsNG.GitHub.Reviews do
...
+        # Ignore dismissed reviews
+        "DISMISSED" -> []
+        # Pass just the username onward
+        id -> [_uid]
+      end
+    end)
+    |> Enum.uniq()
+
     # Remove reviews that don't count
+    %{"CHANGES_REQUESTED" => failed, "APPROVED" => approvals} = reviews
```
umamaistempo: here you could just bind the whole map and then latter add your key :)

```elixir
reviews
|> Enum.flat_map(...)
|> Enum.reduce(...)
|> Map.put("approvers", approvers)
```

lib/github/github/file.ex
```diff
@@ -0,0 +1,41 @@
...
+          "filename" => f
+        }
+      ]) do
+    {:ok,
+     [
+       %BorsNG.GitHub.File{
+         filename: f
+       }
+     ]}
+  end
```
umamaistempo: Okay, looking at the github client call, I think I got it.

You will receive a body like this:

```json
[
  {...},
  {...},
  ...
]
```

The code in this way will always fail because you're doing a pattern that says something like "Receive a list that has a single map as element and that map must have the key 'filename'".

To achieve the objective you want, you should probably map into the results:
```elixir
// [%{}, %{}, %{} ...]
body = Poison.decode!(raw_body)

Enum.map(body, fn el ->
  %BorsNG.Github.File{filename: Map.fetch!(el, "filename"}
end)
```

Also, on the subject of separation of concerns, I think that this module should only receive one "file" per time and the github client should the mapping.

Ie:

```elixir
// this file
def from_json(%{"filename" => filename}),
  do: %BorsNG.Github.File(filename: filename)

// Github client
files =
  body
  |> Poison.decode!()
  |> Enum.map(&Github.File.from_json/1)
{:ok, files}
```

umamaistempo: Note that the Github API _should_ always return the filename, this is an expectation that we have. IMO we should fail hard if this turns out not to be true

lib/github/github.ex
notriddle: This is a toughie, since I think it'll still throw an error in some scenarios, like if the network fails. But, probably, yeah, the fact that it's also a boolean-returning function should probably take precedence, so it would be named `belongs_to_team?`

lib/github/github.ex
notriddle: If it's not, fix it so it is.

lib/github/github/server.ex
```diff
@@ -544,8 +592,12 @@ defmodule BorsNG.GitHub.Server do
     jwt_token = get_jwt_token()
     %{body: raw, status: 201} = "Bearer #{jwt_token}"
     |> tesla_client(@installation_content_type)
-    |> Tesla.post!("installations/#{installation_xref}/access_tokens", "")
+    |> Tesla.post!("installations/#{installation_xref}/access_tokens?members=read", "")
     Poison.decode!(raw)["token"]
+
+#    IO.inspect(raw)
```
notriddle: Please don't commit commented-out code.

lib/github/parse_owner.ex
```diff
@@ -0,0 +1,104 @@
...
+            cond do
+              # glob matches has an extra '/' in it, don't match on it
+             :glob.matches(x.filename, owner.file_pattern) && !:glob.matches(x.filename, owner.file_pattern <> "/*")  ->
+                owner.approvers
+             String.starts_with?(x.filename, owner.file_pattern) ->
+              owner.approvers
+             true ->
+              nil # if unknown fall through
+            end
+          end)
```
notriddle: This looks like a case where you *should* use `flat_map`:

```elixir
          Enum.map(code_owners.patterns, fn owner ->
            cond do
              # glob matches has an extra '/' in it, don't match on it
             :glob.matches(x.filename, owner.file_pattern) && !:glob.matches(x.filename, owner.file_pattern <> "/*")  ->
                [owner.approvers]
             String.starts_with?(x.filename, owner.file_pattern) ->
              [owner.approvers]
             true ->
              [] # if unknown fall through
            end
          end)
```

(the difference, of course, is that in this function, the same analysis is responsible both for filtering and for mapping)

lib/github/parse_owner.ex
```diff
@@ -0,0 +1,104 @@
...
+      owners = %BorsNG.CodeOwners{
+        patterns: []
+      }
+      {:ok, owners}
+    else
+      lines = String.split(file_contents, "\n")
+
+      # Remove any comments from the file
+      lines = Enum.map(lines, fn x ->
+        String.replace(x, Regex.compile!("#.*"), "")
```
notriddle:
```suggestion
        String.replace(x, ~r/#.*/, "")
```

I don't always use regular expressions, but when I do, they're checked at compile-time.

https://elixir-lang.org/getting-started/sigils.html#regular-expressions

lib/worker/batcher.ex
```diff
@@ -528,6 +530,88 @@ defmodule BorsNG.Worker.Batcher do
...
+
+      Logger.info("Passed reviews: #{inspect(passed_review)}")
+
+      # Convert the list of required reviewers into a list of true/false
+      # true indicates that the reviewers requirement was satisfied,
+      # false if it is open
+      approved_reviews = Enum.map(required_reviews, fn x ->
+
+        # Convert a list of OR reviewers into a true or false
+        Enum.reduce(x, false, fn required, acc ->
```
notriddle: `Enum.reduce(x, false, fn _bla, acc -> acc || some_boolish_expr end)` is equivalent to [`Enum.any?(x, fn _bla -> some_boolish_expr end)`](https://hexdocs.pm/elixir/Enum.html#any?/2). In this case, your code can be rewritten to:

```elixir
Enum.any?(x, fn required ->
        # Remove leading @ for team name
        # Split into org name and team name
        team_split = String.slice(required, 1, String.length(required)-1)
                     |> String.split("/")

        # Lookup team ID -> needed later
        {:ok, team} = GitHub.get_team_by_name(repo_conn, Enum.at(team_split, 0), Enum.at(team_split, 1))

        Logger.info("Team: #{inspect(team)}")

        # Loop through reviewers, if they on the team accept their approval
        team_approved = Enum.any?(passed_review["approvers"], fn x ->
          GitHub.is_user_on_team!(repo_conn, x, team.id)
        end)
        Logger.info("Approved: #{inspect(team_approved)}")
        team_approved
    end
end)

File: lib/worker/batcher.ex [hunk 15]
```diff
@@ -528,6 +530,88 @@ defmodule BorsNG.Worker.Batcher do
...
+      missing_code_owners = Enum.with_index(required_reviews)
+                            |> Enum.map(fn {x, i} ->
+
+        if !x do
+          {""}
+        else
+          Enum.at(required_reviews, i)
+        end
+      end)
+                            |> Enum.flat_map(fn x -> x end)
```
notriddle: This is weird indentation and even weirder code. Why not just call `flat_map` up here?

```elixir
      missing_code_owners = Enum.with_index(required_reviews)
                            |> Enum.flat_map(fn {x, i} ->
```

lib/worker/batcher/get_code_owners.ex
```diff
@@ -0,0 +1,18 @@
+defmodule BorsNG.Worker.Batcher.GetCodeOwners do
+  @moduledoc """
+  Get the bors configuration from a repository.
+  This will use `bors.toml`, if available,
```
notriddle: This comment is wrong.

test/parse_owners_test.exs
```diff
@@ -0,0 +1,134 @@
+defmodule BorsNG.ParseTest do
```
notriddle: ❤️ Thank you very much for writing these test cases.

test/batcher/batcher_test.exs
```diff
@@ -311,6 +312,369 @@ defmodule BorsNG.Worker.BatcherTest do
...
+                 "master" => %{".github/CODEOWNERS" =>
+                   ~s"""
+                   *.toml               @my_org/my_team
+                   """},
+               },
+               reviews: %{1 => %{"APPROVED" => 0, "CHANGES_REQUESTED" => 0, "approvers" => []}},
+             }}
+    # When preflight checks reject a patch, no batch should be created!
+    assert [] == Repo.all(Batch)
+  end
```
notriddle: Thank you very much for writing test cases for this part as well. I know what a pain these "big integration tests(tm)" are to work on, since they're running half the code base, but I find that they tend to find bugs.

MarkSRobinson:
@notriddle and @umamaistempo 

Thanks for the feedback.  I'll mark off completed requests with a :zap: sign as a I go.

lib/github/github.ex
MarkSRobinson: ⚡️

lib/github/github/file.ex
MarkSRobinson: ⚡️

lib/github/github/file.ex
MarkSRobinson: ⚡️

lib/github/github/file.ex
MarkSRobinson: ⚡️

lib/github/github/server.ex
MarkSRobinson: ⚡️

lib/github/parse_owner.ex
MarkSRobinson: :zap:

lib/worker/batcher/get_code_owners.ex
MarkSRobinson: ⚡️

test/batcher/batcher_test.exs
MarkSRobinson: They work but they so hard to read and reason about.  You need a really deep understanding of what GitHub is doing to really get it.

lib/github/github.ex
MarkSRobinson: ⚡️

lib/github/github.ex
MarkSRobinson: ⚡️  I've changed everywhere I've added a `bitstring`.  I'm leaving the other bitstrings alone.

lib/github/github/reviews.ex
MarkSRobinson: Gotcha.  ⚡️

lib/github/github/reviews.ex
MarkSRobinson: :zap:

lib/worker/batcher.ex
MarkSRobinson: ⚡️

lib/worker/batcher.ex
MarkSRobinson: So this is something I was playing around with.  Providing the list of missing owners instead of just an error.  I was never super happy about it, so I'll just remove it completely.

lib/github/parse_owner.ex
MarkSRobinson: In this case I don't think so.

I need to do some odd filtering.  A file can only have one required reviewer and the last match takes precedence.  So I need to keep the reviewer list together until later.

I've added some comments discussing this.

lib/github/github/reviews.ex
MarkSRobinson: ⚡️

lib/github/github/file.ex
MarkSRobinson: Thanks ⚡️

MarkSRobinson:
@notriddle and @umamaistempo

I've added the change you have suggested.

notriddle:
@MarkSRobinson How's the progress on this going. The RFC is approved, and all our issues have been resolved. It still has the `[wip]` label on it, though. Would you say it's ready to take it off? Or is there something else holding us up?

MarkSRobinson:
@notriddle Thanks for reminding me.  It's ready to land.

PR #727, opened 2019-08-18
delegate: allow abbreviation to d

### LINKED ISSUE #615: Shorten delegate to d?
I find that I use `bors delegate+` a lot: almost every PR has some minor issues, but most of PRs have *only* minor issues, so "r=me with nits fixed" is a common outcome of the review. `bors delegate+` is very handy in this situation: you don't need to review PR the second time, and you give a strong signal of trust to the contributor.

The only problem is that delegate is a long word to type, so I feel stressed when I try to type it without misspellings :) 

I wish I could

bors d+

Given that bors replies with a helpful message, I think it's OK that d, by itself, is not self-explanatory?

tommilligan:
Closes #615.

Add the new commands:
- `d+` (alias of `delegate+`)
- `d=` (alias of `delegate=`)

Add one new dev/test dependency, [`ex_parameterized`](https://github.com/KazuCocoa/ex_parameterized).


notriddle:
Okay, I've asked for some feedback on the forum. If nobody says anything in a couple weeks, we'll probably merge it.

PR #729, opened 2019-08-21
Include child team members in collaborators

indirect:
So I'm running into a problem where Bors sync isn't showing collaborators that I would be expecting it to show. The deciding factor _seems_ to be that this organization is very into nested teams, and anyone who is in the top-level team shows up, but anyone in a tested team does not show up. That plus GitHub's [Collaborator API docs][1] lead me to believe that Bors needs to use the `hellcat-preview` content type to get nested team members when syncing collaborators.

Anyone else run into this? Concerns, other thoughts?

[1]: https://developer.github.com/v3/repos/collaborators/


indirect:
Update: I deployed this to Heroku, it fixes everything for me. :tada:

PR #737, opened 2019-09-07
Do pagination on the History/Log page

### LINKED ISSUE #305: Do pagination on the History/Log page
The History page (in the `project_controller.ex` file under `def log`, and the `log.html.eex` file) shows the entire history of the given project when you first open it. It should probably have a limit on that, with an Older button on the bottom to page through it (or infinite scrolling, if you really feel fancy).

kimsaehun:
Fixes #305 

I added an "Older" button at the bottom of the page that will retrieve older logs from the database.

GitHub alerted me that the branches could not be automatically merged but I've done a git rebase which I believe was performed correctly.


notriddle:
It works right when I try it.

PR #753, opened 2019-09-21
Adding support for asterisk on CodeOwners

gmendonca:
Follow up from https://github.com/bors-ng/bors-ng/pull/725

There were some logic missing for CodeOwners check, namely:
- Global fallback with `*`
- Two consecutive asterisks `**` in patterns matched against full pathname

I did a super simple approach, not sure if it's the best one. Feedbacks appreciated.

### DIFF
--- a/lib/github/parse_owner.ex
+++ b/lib/github/parse_owner.ex
@@ -44,7 +44,7 @@ defmodule BorsNG.CodeOwnerParser do
               owner.approvers
              String.starts_with?(x.filename, owner.file_pattern) ->
               owner.approvers
-            true ->
+             true ->
               nil # if unknown fall through
             end
           end)
@@ -60,7 +60,7 @@ defmodule BorsNG.CodeOwnerParser do
             end
           end)
       end)
-        
+ 
     required_reviewers = Enum.filter(required_reviewers, fn x -> x != nil end)
 
     Logger.debug("Required reviewers: #{inspect(required_reviewers)}")
--- a/lib/github/parse_owner.ex
+++ b/lib/github/parse_owner.ex
@@ -75,7 +75,7 @@ defmodule BorsNG.CodeOwnerParser do
       String.starts_with?(file_pattern, double_asterisk) ->
         pattern = String.trim_leading(file_pattern, double_asterisk)
         String.contains?(file_name, pattern)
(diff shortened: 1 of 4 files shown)


lib/github/parse_owner.ex
```diff
@@ -56,14 +60,30 @@ defmodule BorsNG.CodeOwnerParser do
...
 
+  @spec process_double_asterisk(String.t, String.t) :: boolean
+  def process_double_asterisk(file_name, file_pattern) do
+    double_asterisk = "**"
+    cond do
+      String.starts_with?(file_pattern, double_asterisk) ->
+        pattern = String.trim_leading(file_pattern, double_asterisk)
+        String.contains?(file_name, pattern)
+      String.ends_with?(file_name, double_asterisk) ->
+        pattern = String.trim_leading(file_pattern, double_asterisk)
```
notriddle: Wait, it's really checking the `file_name` for a "`**`"? I can't find anything about it in GitHub's documentation. It only mentions asterisks as being special in patterns.

gmendonca: Accordingly to https://help.github.com/en/articles/about-code-owners: `A CODEOWNERS file uses a pattern that follows the same rules used in gitignore files.` https://git-scm.com/docs/gitignore#_pattern_format shows how the pattern support is for `**`.

gmendonca: That being said, I did a very simple logic that I'll try to explain. For example, if the pattern is `abc/**` I remove the `**` asterisk and check if the file name starts with `abc/`. The other way around if it's `**/foo`, I remove the `**` and check if the file name has `foo` on it. Did I managed to explain correctly? Not super sure if it's the best approach. I'll be glad to change if you have a better idea.

notriddle: Yes, but that's for the *patterns*. `String.ends_with?(file_name, double_asterisk)` checks whether the file's own name ends with double asterisks.

gmendonca: Oh, wait, but in this part I should be removing the trailing one...

gmendonca: I added tests for all the cases described as a proof of concept

gmendonca: I totally misunderstood you, I'm sorry, you are correct, just fixed it.

PR #758, opened 2019-09-24
Add gitpod.io config files

notriddle:
See https://www.gitpod.io/blog/gitpodify/


umamaistempo:
Niiice, really liked it. Will try to gitpod it locally here and, if it works, LGTM :+1:

notriddle:
Yes, you're right, the test script was messed up.

Here's a new commit that fixes it when I tried it.

Issue #754, opened 2019-09-21
Batch with PRs with different CodeOwners lead to crash


gmendonca:
Situation:
 - PR 1 has files modified that team 1 is the code owner of them
 - PR 2 has files modified that team 2 is the code owner of them
 - When batched together the batch fails saying that with the 422 code (related to https://github.com/bors-ng/bors-ng/issues/674) that the team 1 hasn't  approved.
 - However if I merge them separately with bors, it goes without a problem

My hypothesis is maybe Github fails to recognise the rightful code owner for each PR and instead thinks of them as a single thing and then check each PR looking for the code owner of all files modified.

I don't think https://github.com/bors-ng/bors-ng/issues/681 will help in this scenario as both will get reschedule in the same batch again.

MarkSRobinson:
I've hit this problem before.  The solution it to turn off the GitHub review requirements, and enforce it solely through bors w/ https://github.com/bors-ng/bors-ng/pull/725 applied.

gmendonca:
Uhmmm, @MarkSRobinson  after merging https://github.com/bors-ng/bors-ng/pull/718 I cannot use both Squash merge and Codeowners options together. And I think what you just said should solve both problems. I'll give it a try.

gmendonca:
Closing it, as solely relying on bors solved the problem. I'll update bors documentation as well.

PR #762, opened 2019-10-01
Batch per code owner

gmendonca:
Working to try to fix the issues presented on: https://github.com/bors-ng/bors-ng/issues/754

I was thinking to avoid batching together patches with different code owners. Not sure if it's the best approach, so feedbacks are welcome here.

### DIFF
--- a/lib/database/code_owner.ex
+++ b/lib/database/code_owner.ex
@@ -1,29 +1,29 @@
 defmodule BorsNG.Database.CodeOwners do
-  
-    use BorsNG.Database.Model
-  
-    @type t :: %__MODULE__{}
-    @type id :: pos_integer
-  
-    schema "code_owners" do
-      field :name, :string
-    end
-  
-    @spec changeset(t | Ecto.Changeset.t, map) :: Ecto.Changeset.t
-    @doc """
-    Builds a changeset based on the `struct` and `params`.
-    """
-    def changeset(struct, params \\ %{}) do
-      struct
-      |> cast(params, [
-        :name])
-      |> unique_constraint(:name, name: :code_owners_name_index)
-    end
 
-    @spec all_for_patch(Patch.id) :: Ecto.Queryable.t
-    def all_for_patch(patch_id) do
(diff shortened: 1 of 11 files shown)


notriddle:
You know what?

How often do you expect two patches to have the same code owner? If the answer is "not very often", maybe we should just disable batching in the presence of a `CODEOWNERS` file? It would be a lot simpler.

gmendonca:
In my situation, sometimes we are trying to merge 10-12 PRs with different code owners and they all are cascading failing because of 422 errors.

What I normally do in this scenario is: I manually separate PRs with the same `CODEOWNERS` with the same priority using `bors r+ p=`. However this is a very dull and manual process.

notriddle:
That's not really my question.

My question is: since we're already separating them by priority, and by branch name, by mergeability, and now separating them by code owner. With all these different things forcing separate batches, are we still going to perform enough batching to make it worthwhile having it at all? I'm wondering if it would make any difference if we just disabled batching entirely instead of trying to be smart about it.

gmendonca:
Yeah, you got a really good point. 

Normally I would expect, in the `CODEOWNERS` situation, to have `* @my_org/team_name` in which most of the files would fall into, and some more restrictives rules after it. I would expect to have more PRs being batched together that falls into the general rule, and few more restrictives ones not being batched together.

But again, you got a really good point. Since `bors` is done to be generic, in a sense, if someone is using `CODEOWNERS` in a different way, it would potentially lead to not batching PRs at all.

notriddle:
> Normally I would expect, in the CODEOWNERS situation, to have * @my_org/team_name in which most of the files would fall into, and some more restrictives rules after it.

If that's what you're doing, then *awesome*, my concerns have been answered 🎉 . We can continue this PR with confidence that it will actually be useful to someone.

gmendonca:
Closing as the solutions is to only rely on bors to validate if there's a code owner approval or not. Letting both Github and bors with this enable lead to problems described here: https://github.com/bors-ng/bors-ng/issues/754

PR #763, opened 2019-10-03
Potentially fixing situations where everything works, but when pushing

gmendonca:
some errors like 422 might happen. Creating as a draft to get feedbacks since it isn't tested.


gmendonca:
Also closing in favor of https://github.com/bors-ng/bors-ng.github.io/pull/92

Issue #766, opened 2019-10-11
`delete_merged_branches` doesn’t work with `use_squash_merge`


EFanZh:
I have

```toml
delete_merged_branches = true
use_squash_merge = true
```

in my `bors.toml` file.

After a PR being merged, the source branch in the same repository still exists.

PR #778, opened 2019-10-18
Fixing branch deletion when PR is closed when using squash and merge

gmendonca:
Fix https://github.com/bors-ng/bors-ng/issues/766

### DIFF
--- a/lib/worker/branch_deleter.ex
+++ b/lib/worker/branch_deleter.ex
@@ -72,7 +72,7 @@ defmodule BorsNG.Worker.BranchDeleter do
 
     pr_squash_merged = String.starts_with?(pr.title, "[Merged by Bors] - ")
 
-    if pr_in_same_repo && delete_merged_branches && (pr.merged || pr_closed) do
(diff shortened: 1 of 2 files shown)


lib/worker/branch_deleter.ex
```diff
@@ -66,8 +68,19 @@ defmodule BorsNG.Worker.BranchDeleter do
...
     end
 
-    if pr_in_same_repo && delete_merged_branches && pr.merged do
-      GitHub.delete_branch!(conn, pr.head_ref)
+    pr_closed = pr.state == :closed
+
+    pr_squash_merged = String.starts_with?(pr.title, "[Merged by Bors] - ")
+
+    if pr_in_same_repo && delete_merged_branches do
+      cond do
```
gmendonca: Really not sure if it's the best way to do it, seemed more clear to me though.

notriddle: Seems fine to me.

notriddle:
Let's do this, then.

bors r+

PR #779, opened 2019-10-21
Format generated commit message

EFanZh:
Subject and body in commit message should be separated with a blank line.

For example, if the title is:

```
Title
```

and the body is

```
Paragraph 1 

Paragraph 2
```

Then the generated commit message should be:

```
Title

Paragraph 1 

Paragraph 2
```

not:

```
Title
Paragraph 1 

Paragraph 2
```

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -357,7 +357,7 @@ defmodule BorsNG.Worker.Batcher do
               %{
                 tree: merge_commit.tree,
                 parents: [prev_head],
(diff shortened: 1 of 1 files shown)


notriddle:
You're right. I'm surprised I messed that up.

bors r+

notriddle:
Uh, yeah, that's not going to work until #780 merges.

bors r-

notriddle:
Okay, let's try that again.

bors r+

PR #780, opened 2019-10-21
update taskcluster.yml with community ci provisionerId/workerType

milescrabill:
- this should work after switching the taskcluster github integration over to `community-tc-integration`.
- the current integration will continue to work through November 9th, at which point this change will be needed

cc @djmitche


milescrabill:
> Submitting the task to Taskcluster failed. Details

Note that this failure is because the integration hasn't been updated, the workerType and provisionerId changes are only scoped on the community deployment of Taskcluster. (so this is expected)

notriddle:
Okay, I've updated the integration. Let's try it again:

bors try

notriddle:
Awesome.

bors r+

Issue #430, opened 2018-06-18
Provisionally accept 'r+' before PR status satisfied


jeffparsons:
Hi there! 👋 

One of the really nice things we've found about using Bors is that once a reviewer is happy with a PR's code changes, they can 'r+' it, and then (assuming all goes well) some time later `master` will contain those changes. No "babysitting PRs" involved.

We recently started using the `pr_status` option to require branches to pass a restricted set of basic tests before they are admitted into a Bors batch, in an attempt to reduce the high number of batches that people were accidentally poisoning.

This has helped overall, but the way `pr_status` currently works means that we're now babysitting PRs again, like we were before adopting Bors. I.e. if we "r+" a PR before its own tests have passed, Bors tells us:

> 👎 Rejected by PR status

So someone has to wait for the branch's own tests to pass, and then "r+" it.

**Proposal:** It seems like a more natural workflow would be to let you "r+" a PR, and then as soon as it meets all other criteria (e.g. statuses, GitHub reviews, etc.), _then_ Bors would admit it into a batch. And GitHub does appear to offer status update events, so it wouldn't require Bors to poll GitHub.

Is there some reason I've missed why rejecting the "r+" is sometimes preferable?

(We've actually now written a bot internally to approximate this proposed behaviour — but I'd love to be able to delete our hacks if this behaviour would be preferred upstream. 😁)

JensRantil:
Duplicate of #390?

jeffparsons:
@JensRantil Ah, yes, I suppose it is — at least in spirit. I missed that one. I could move the text of this issue into a comment on https://github.com/bors-ng/bors-ng/issues/390?

JensRantil:
@jeffparsons If it adds additional information to what's already written in #390, sure! If not, we can close this.

jeffparsons:
Closing in favour of https://github.com/bors-ng/bors-ng/issues/390

Issue #390, opened 2018-04-03
Auto-merge onGithub approval+`bors r+`


JensRantil:
We are using `required_approvals = 1`. Scenario:

 1. Author submits PR which is a simple documentation fix fine to merge immediately as a peer has approved it.
 2. The author write `bors r+` that I'm fine with it being merged as soon as someone Github-approves the PR.
 3. The peer approves the PR using Github review.

Expected:
 * That Bors tells me in a comment that it will merge the PR immediately as soon as enough people have reviewed/approved it.
 * That the PR is merged by Bors immediately after 3.

What happens:

 * Bors tells me it needs approval after 2.
 * I, as the author, must go into the PR again after 3 and write `bors r+`. This adds latencies/energy to quickly get small fixes like this into the main branch.

kamalmarhubi:
A big part of the draw of bors is that it makes merging PRs less annoying, in addition to safer. This seems really worthwhile as an improvement.

jeffparsons:
Originally posted here: https://github.com/bors-ng/bors-ng/issues/430. Then it was pointed out that it was basically a dup of this ticket.

----

Hi there! 👋 

One of the really nice things we've found about using Bors is that once a reviewer is happy with a PR's code changes, they can 'r+' it, and then (assuming all goes well) some time later `master` will contain those changes. No "babysitting PRs" involved.

We recently started using the `pr_status` option to require branches to pass a restricted set of basic tests before they are admitted into a Bors batch, in an attempt to reduce the high number of batches that people were accidentally poisoning.

This has helped overall, but the way `pr_status` currently works means that we're now babysitting PRs again, like we were before adopting Bors. I.e. if we "r+" a PR before its own tests have passed, Bors tells us:

> 👎 Rejected by PR status

So someone has to wait for the branch's own tests to pass, and then "r+" it.

**Proposal:** It seems like a more natural workflow would be to let you "r+" a PR, and then as soon as it meets all other criteria (e.g. statuses, GitHub reviews, etc.), _then_ Bors would admit it into a batch. And GitHub does appear to offer status update events, so it wouldn't require Bors to poll GitHub.

Is there some reason I've missed why rejecting the "r+" is sometimes preferable?

(We've actually now written a bot internally to approximate this proposed behaviour — but I'd love to be able to delete our hacks if this behaviour would be preferred upstream. 😁)

milesgould:
[...]

PR #785, opened 2019-10-29
Wait for CI to finish instead of rejecting immediately on bors r+ when CI is still "pending"

zli-simspace:
Hi!

Our company uses Bors and I've made these changes (for our own use) that we thought could also be useful upstream.

## Problem

This issue describes the motivation for this PR pretty well:

https://github.com/bors-ng/bors-ng/issues/430

The gist of it that we want to mark a PR as ready to merge before CI has finished building and instead of an immediate

> :-1: Rejected by PR status

we'd like Bors to wait until CI is finished.

## Solution

We distinguish between Github statuses that are errors and waiting differently (instead of treating them all as a failed `passed_status`. In the case where there are no errors and some waiting statuses, notify with a response and wait polling with an exponential backoff until we're in some other state than "no errors and some waiting statuses" (i.e., `patch_preflight` returns something else).

Once that happens, report back and take action as Bors did before.

Waiting on Github's waiting status can timeout, which is considered a failure and Bors won't merge the PR. This a timeout can be set in `bors.toml` as `prerun_timeout_sec`. The default is 30 minutes.

## Tests

I've added two tests for the new functionality.

## Relevant issues

https://github.com/bors-ng/bors-ng/issues/430

Though #430 was closed in favour of #390, a timeout makes much more sense for #430. Our specific pain point does not relate to Github approvals and was really about the issue with

- waiting for CI (an automatic process),
- coming back to type `bors r+` (manual) and then
- have CI build the `staging` branch (another automatic process.


lib/web/controllers/project_controller.ex
```diff
@@ -331,9 +331,13 @@ defmodule BorsNG.ProjectController do
             {:ok, "Successfully added #{user.login} as a reviewer"}
         end
     end
-    conn
-    |> put_flash(state, msg)
-    |> redirect(to: project_path(conn, :settings, project))
+    case conn do
+      nil -> msg
```
notriddle: Why is `conn` winding up as `nil` here? This seems like a sign that there's another bug outside of this code.

zli-simspace:
Great, thanks for the quick initial response on this PR!

lib/web/controllers/project_controller.ex
zli-simspace: Oops, thanks for catching this. That's a change I made for easier development that I forgot to undo. Just pushed a change to put this back the way it was.

For context, though not needed for this PR, I used a "friendlier" version of ServerMock (in conjunction with the REPL in `iex -S mix phx.server`) to get a better idea of how Bors works and to do development. I had trouble getting the nested maps on ServerMock in the right state otherwise. https://github.com/bors-ng/bors-ng/pull/785/commits/babca4b78908b058451a554a08877d096a3c10c8

notriddle:
Okay, [here's the RFC](https://forum.bors.tech/t/wait-for-ci-to-finish-instead-of-rejecting-immediately-on-bors-r-when-ci-is-still-pending/388) with all the draft documentation for this. If it's wrong, you should be able to log into the forum (with your github account) to edit it.

lib/web/controllers/project_controller.ex
notriddle: That's pretty cool. If you still want it, you can open another PR for it. I'd be happy to accept your `FriendlyMock` system into bors for developer convenience.

zli-simspace: Cool, will do!

lib/worker/batcher.ex
```diff
@@ -189,6 +186,25 @@ defmodule BorsNG.Worker.Batcher do
     end
   end
 
+  def handle_info({:prerun_poll, timeout, args}, proj_id) do
+    {reviewer, patch} = args
+    project = Repo.get(Project, patch.project_id)
+    repo_conn = get_repo_conn(project)
+    {:ok, toml} = Batcher.GetBorsToml.get(repo_conn, patch.commit)
+    prerun_timeout = toml.prerun_timeout_sec * 1000
```
bbuchalter:
```suggestion
    prerun_timeout_ms = toml.prerun_timeout_sec * 1000
```

lib/worker/batcher.ex
```diff
@@ -189,6 +186,25 @@ defmodule BorsNG.Worker.Batcher do
...
+    case patch_preflight(repo_conn, patch) do
+      :ok ->
+	run(reviewer, patch)
+      :waiting when timeout > prerun_timeout ->
```
bbuchalter:
```suggestion
      :waiting when timeout > prerun_timeout_ms ->
```

lib/worker/batcher/bors_toml.ex
```diff
@@ -15,6 +15,9 @@ defmodule BorsNG.Worker.Batcher.BorsToml do
 
   defstruct status: [], block_labels: [], pr_status: [],
     timeout_sec: (60 * 60),
+    # Half an hour. Give up when more time is spent waiting for this amount
+    # between two successive polls.
```
bbuchalter: This comment is confusing to me. Based on my reading of the PR, maybe this would be better?
```
prerun_timeout_sec controls how long bors will wait for all GitHub status checks to be completed before taking action. If this value is set to 0, bors will not wait for status checks to be completed. If all status checks are not completed within prerun_timeout_sec, bors will return an error message.
```

lib/worker/batcher/message.ex
```diff
@@ -28,6 +28,15 @@ defmodule BorsNG.Worker.Batcher.Message do
     {"Delayed for higher-priority pull requests", :running}
   end
 
+  def generate_message({:preflight, :waiting}) do
+    ":clock1: Waiting for PR status (Github check) to be set, probably by CI. Bors will automatically try to run when all required PR statuses are set."
```
bbuchalter: Two changes I'd like to see here:
* Let's use GitHub's language: they call these things [status checks](https://help.github.com/en/github/administering-a-repository/enabling-required-status-checks)
* It'd be _great_ if this message explained how long Bors is configured to wait until erroring. We'd need access to `prerun_timeout_sec` - not sure if we can get that here easily.

```
    ":clock1: Bors is waiting for all GitHub status checks to complete and will continue polling until $TIME_IN_FUTURE
```

lib/worker/batcher/message.ex
```diff
@@ -28,6 +28,15 @@ defmodule BorsNG.Worker.Batcher.Message do
...
+  end
+  def generate_message({:preflight, :ok}) do
+    "All preflight checks passed. Batching this PR into the staging branch."
+  end
+  def generate_message({:preflight, :timeout}) do
+    "Timeout waiting for PR status (Github check). Likely CI was taking too long."
```
bbuchalter:
```suggestion
    "GitHub status checks took too long to complete, so bors is giving up. You can adjust bors configuration to have it wait longer if you like."
```

zli-simspace:
@bbuchalter thanks for your improvements! They look good to me.

lib/worker/batcher/bors_toml.ex
zli-simspace: Yeah, I think your suggestion is good.

> If all status checks are not completed within prerun_timeout_sec,

The only thing to note is that I was lazy and the actual timeout is proportional to prerun_timeout_sec but isn't prerun_timeout_sec. prerun_timeout_sec is the amount the last exponential backoff needed to wait.

You can get one timeout from the other by calculating the difference of two geometric series (or the inverse of that).

It was done out of laziness and to have fewer moving parts. But feel free to suggest an improvement to that as well (ideally one that isn't too complicated).

dchaley:
➕  to this being a very useful PR that captures the promise of a smart, deferred merge queue. Anything blocking merge?

lib/worker/batcher/bors_toml.ex
bbuchalter: Lazy is fine, just so long as it's clear. I think we just need to explain it with words so folks understand their configuration choices.

notriddle:
Nothing's really blocking it at this point, no.

zli-simspace:
Oh, I thought I had to wait for the RFC to close. Should I be r+ing it myself?

notriddle:
I was waiting for the RFC to close, yes. It'll literally be closed in a day, and nobody's said anything, so I'll probably merge it then.

phanle:
Can't wait to see this feature landed. This would reduce greatly context switches for our team at Gusto!

pksunkara:
This is not perfect.

First of all, I don't think any message is being sent when timeout happens (default was 30m). As you can see [here](https://github.com/clap-rs/clap/pull/1657), the CI took 1 hr to finish and for 3 hrs after that, nothing happened from bors. I had to `r+` again.

Secondly, why can't we use the events from Github status checks to run bors instead of doing this polling and timing out stuff?

notriddle:
The lack of message on failure is bad. The fact that it relies exclusively on polling is implementation laziness.

The fact that it has a timeout is a necessary feature of writing good network software; we don't want to wait forever for a message that will never come. This is also why I would prefer to have polling *and* webhook delivery running alongside each other.

I'm fine with merging imperfect implementation code, as long as it's basically usable and can be improved without introducing any breaking changes.

pksunkara:
We can wait for the webhook and poll just once around the timeout. That would reduce the polling and still be network fault-tolerant.

notriddle:
Yeah, I agree, that would be better.

pksunkara:
In fact, I want to incorporate more github workflow into bors (with options of course). Let me write a quick RFC.

PR #807, opened 2019-11-21
Fixes in CODEOWNERS for patterns without a leading slash

zegl:
As per the .gitignore pattern format [1]
The pattern "foo/" matches both "foo/bar.txt" and "somewhere/deep/foo/hello-world.txt"
The pattern "foo/**/bar/*.txt" matches both "foo/a/bar/b.txt" and "somewhere/deep/foo/a/bar/b.txt"

This patch adds special treatment for patterns that start with a leading slash, and for patterns that doesn't start with a leading slash.

1: https://git-scm.com/docs/gitignore#_pattern_format


zegl:
I'd like to get a second pair of eyes on this patch, to make sure that the behaviour is as expected.

PR #823, opened 2019-12-13
Use destination branch name in squash+merge comment

ChristianAlexander:
This change updates batcher to comment with the name of the branch the merge was performed in, rather than hard-coding `# Pull request successfully merged into master.`.

refactor: produce squash merge message in message.ex
feat: comment correct branch name in squash merge

PR #832, opened 2019-12-20
Rename PR after sending all the comments

phanle:
Context: 
Changing the PR title then adding comments to the PR would trigger email clients like Gmail to think it is a different thread in email notifications.

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -524,16 +524,17 @@ defmodule BorsNG.Worker.Batcher do
     |> Patch.all_for_batch()
     |> Repo.all()
 
+    send_message(repo_conn, patches, {:succeeded, statuses})
+
     if toml.use_squash_merge do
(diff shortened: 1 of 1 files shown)


lib/worker/batcher.ex
```diff
@@ -524,16 +524,17 @@ defmodule BorsNG.Worker.Batcher do
     |> Patch.all_for_batch()
     |> Repo.all()
 
+    send_message(repo_conn, patches, {:succeeded, statuses})
+
     if toml.use_squash_merge do
       Enum.each(patches, fn patch ->
+        GitHub.post_comment!(repo_conn, patch.pr_xref, "# Pull request successfully merged into master.")
```
notriddle: It doesn't look like you properly rebased this onto https://github.com/bors-ng/bors-ng/pull/823

phanle: 👍 fixing now

PR #834, opened 2019-12-24
Add Batch details page that links from Repository history tab

noizwaves:
Hey, we've been using bors-ng and our developers were having trouble finding the corresponding CI build for their batched PRs. We run a lot concurrent builds, so this can be time consuming and very manual.

In this PR we add a Batch details page, which shows some basic information about the batch, as well as the batch's Status links (if present). This allows a developer to discover their corresponding CI build (`buildkite/shipit-test` in the attached screenshot).

Basic Batch details page

To get to this page, we added a link from the Repository history tab. We weren't sure of the best way to add this, so we converted the existing "Batch" text into a "Batch X" hyperlink to the details page (see `Batch 4` in attached screenshot).

History tab containing link

We have other ideas we'd like your feedback/thoughts on too. Maybe we can update the bors status url we send to the GitHub Status API to point to the new details page (instead of the History page)?

### DIFF
--- a/lib/web/templates/batch/show.html.eex
+++ b/lib/web/templates/batch/show.html.eex
@@ -1,39 +1,29 @@
 <main role=main><div class=wrapper>
 
-<h2 class=header>Batch details</h2>
+<h2 class=header>Batch Details</h2>
 
-<table id="batch-details-table" class="table">
-  <tbody>
-    <%= if @batch do %>
-      <tr>
-        <td>State</td>
-        <td><%= @batch.state %></td>
-      </tr>
-      <tr>
-        <td>Priority</td>
-        <td><%= @batch.priority %></td>
-      </tr>
-
-      <tr>
-        <td>Pull requests</td>
-        <td>
-          <%= for patch <- @patches do %>
-            <a href="<%= Confex.fetch_env!(:bors, :html_github_root) %>/<%= @project.name %>/pull/<%= patch.pr_xref %>">#<%= patch.pr_xref %></a>
-          <% end %>
-        </td>
-      </tr>
-      <tr>
-        <td>Statuses</td>
-        <td>
-          <%= for status <- @statuses do %>
-            <a href="<%= status.url %>"><%= status.identifier %></a>
-          <% end %>
-        </td>
-      </tr>
-  <% else %>
-    There is no such batch
+<%= if @batch do %>
+<p>Priority: <%= @batch.priority %></p>
+<p>State: <%= stringify_state(@batch.state) %></p>
+<p>
+  Status:
+  <%= for status <- @statuses do %>
+    <%= if status.url do %>
+    <a href="<%= status.url %>"><%= status.identifier %></a>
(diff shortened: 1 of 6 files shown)


lib/web/templates/batch/show.html.eex
```diff
@@ -0,0 +1,29 @@
...
+<p>Priority: <%= @batch.priority %></p>
+<p>State: <%= stringify_state(@batch.state) %></p>
+<p>
+  Status:
+  <%= for status <- @statuses do %>
+    <%= if status.url do %>
+    <a href="<%= status.url %>"><%= status.identifier %></a>
+    <% else %>
+    <span><%= status.identifier %></span>
+    <% end %>
```
notriddle: One thing that seems like it would be really useful here would be to show the state of individual statuses. These are recorded in the database, so it should be easily doable by adding a line like this to the end:

```iex
(<%= stringify_state(status.state) %>)
```

So that it'll report something like `ci/buildkite (Running)`

notriddle:
This looks awesome! It's basically perfect right out of the shoot.

> In this PR we add a Batch details page, which shows some basic information about the batch, as well as the batch's Status links (if present). This allows a developer to discover their corresponding CI build (buildkite/shipit-test in the attached screenshot).

- [ ] I've left a comment on your template, because there's some info I think you're missing. Overall, it's the right idea.

> To get to this page, we added a link from the Repository history tab. We weren't sure of the best way to add this, so we converted the existing "Batch" text into a "Batch X" hyperlink to the details page (see Batch 4 in attached screenshot).

- [ ] That's fine. I would also add a link somewhere on the `Pull requests` tab, so that we could easily go from the currently running PR to its batch status page, as in `<h1><a href="/batches/{n}">Running</a></h1>` or something.

> We have other ideas we'd like your feedback/thoughts on too. Maybe we can update the bors status url we send to the GitHub Status API to point to the new details page (instead of the History page)?

Sure, as long as the batch status page has all the information as the history line, it's fine.

- [ ] If you're going to do that, add the date to the page. That's the only piece of information in the table that's not in the page.

ansonlc:
@notriddle thanks for the feedback! I added status state. We can change the status link (also add the date) later in a separate PR.
Happy holidays!

lib/web/templates/batch/show.html.eex
```diff
@@ -9,9 +9,9 @@
   Status:
   <%= for status <- @statuses do %>
     <%= if status.url do %>
-    <a href="<%= status.url %>"><%= status.identifier %></a>
+    <a href="<%= status.url %>"><%= status.identifier %> (<%= stringify_state(status.state) %>)</a>
     <% else %>
-    <span><%= status.identifier %></span>
+    <span><%= status.identifier %> (<%= stringify_state(status.state) %>)</span>
```
ansonlc: assuming the state is not null

notriddle:
Cool, I'll go ahead and merge this, then.

bors r+

PR #839, opened 2020-01-02
Add a configuration option for batch size limiting

### LINKED ISSUE #205: Add a configuration option for batch size limiting
There are really two possible ways that you want to limit the size of a batch:

* Patch-specific limits. This would be used by servo/servo (that need to sync up with a worlds-apart Mercurial repo).
* Repo-wide limits. rust-lang/rust would want this for their bisect process, and @mmun would want it since their company doesn't want batching at all.

kimsaehun:
Fixes #205 

Took a while but I've implemented a 'r+ single on/off' command to have patches be batched by themselves only. However, this only occurs when patches are batched, either during bisect or when a patch is newly reviewed. I couldn't figure out how to make this apply for patches that are already in a running/waiting batch.

Additionally, I wasn't too sure whether ensuring that bisect honors the single patch option was a good idea or not. I'm thinking that it's unnecessary because if a batch fails, the singled patch will either become singled through enough bisections or will successfully merge into master without any issues. I implemented it just in case there was a case I have not thought of but please let me know if it can be removed. 

I've added a few test cases to test for the following:

- patches that are newly reviewed and singled are batched alone and not into existing waiting batches
- patches that are newly reviewed are not added to batches containing singled patches
- the bisecting process honors single patches

### DIFF
--- a/lib/database/patch.ex
+++ b/lib/database/patch.ex
@@ -21,6 +21,7 @@ defmodule BorsNG.Database.Patch do
     field :commit, :string
     field :open, :boolean, default: true
     field :priority, :integer, default: 0
+    field :is_single, :boolean, default: false
     belongs_to :author, User
     timestamps()
   end
@@ -40,7 +41,8 @@ defmodule BorsNG.Database.Patch do
       :author_id,
       :open,
       :into_branch,
-      :priority])
+      :priority,
+      :is_single])
     |> unique_constraint(:pr_xref, name: :patches_pr_xref_index)
   end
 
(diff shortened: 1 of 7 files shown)


notriddle:
Since this PR adds a new command, it should probably [go through an RFC first](https://forum.bors.tech/t/about-the-draft-rfcs-category/291). Basically, open a topic on the forum, then wait for a month for feedback.

kimsaehun:
Thanks for the heads up! I wrote one [here](https://forum.bors.tech/t/allow-patches-to-be-batched-by-themselves-only/425).

notriddle:
Rebased onto master and added the standalone `single` command.

bors r+

notriddle:
bors r+

Busted merge. Trying again.

PR #843, opened 2020-01-09
Isolate unmergeable PRs when a batch fails because of conflict

noizwaves:
## Context
In our deployment of bors, we see numerous conflicts within batches. Batches often contain PRs that are not mergeable with the target branch. These batches will sit in `:waiting` state, collecting other PRs that have been added to the merge queue. These unmergeable PRs will remain in the merge queue, creating failures in many batches and slowing the progress of mergeable PRs through the merge queue.

This problem was briefly mentioned in [this forum post](https://forum.bors.tech/t/limiting-the-scope-of-merge-failures-in-batches/421).

## Goal
We want to remove unmergeable PRs from the merge queue as soon as possible. We want to give the authors quicker feedback that they need to manually rebase their branch. Isolating unmergeable PRs to their own batches means when they do fail, they will not be bisected/retried.

## Approach
Currently bors performs a bisect in a merge conflict situation to identify the problematic PRs. We think that by using additional information (the mergeability of PRs with the target branch), we can perform more optimal identification.

## Solution
When a batch fails due to a merge conflict, we isolate any PRs that are unmergeable, and put them into their own individual batches. The goal is that these unmergeable batches will run immediately (after the current batch finishes) and fail fast. This contains the failure to only this PR, and does not affect other PRs.

We use the GitHub API to determine mergeability (using the `mergeable` field referenced [here](https://developer.github.com/v3/pulls/#response-1)).

## Example
Here is an example of the proposed solution. 

Screen Shot 2020-01-09 at 5 33 29 PM

The timeline of events
1. PRs 97, 93, 91, 90, and 89 are all added to the merge queue. 97, 92, and 89 have merge conflicts with the target branch and are unmergeable.
1. The 5 PRs enter Batch 138
1. Batch 138 fails due to a merge conflict
1. bors first determines that PRs 97, 92, and 89 are unmergeable and creates individual batches for each of them (Batch 139, Batch 140, and Batch 141)
1. then bors creates a batch containing the remaining mergeable PRs 90 and 91 (Batch 142)
1. bors runs Batch 139 and fails due to a merge conflict. This repeated for Batches 140 and 141 also.
1. bors runs Batch 142 and succeeds.

## Possible extensions
- fail unmergeable batches immediately (skip the queue by creating with state = `:failed`)?
- bump priority of unmergeable batches?
- perform best effort merging  when there are conflicts within a batch only (not with target). Instead of failing and bisecting when a conflict occurs, isolate the temporary conflict patches and proceed with remaining patches.

### DIFF
--- a/lib/github/github/friendly.ex
+++ b/lib/github/github/friendly.ex
@@ -161,7 +161,8 @@ defmodule BorsNG.GitHub.FriendlyMock do
              head_sha: sha,
              head_ref: ref,
              body: body,
-             user: @def_user}
+             user: @def_user,
+             mergeable: true}
     update_mock([:pulls], &(Map.put(&1, number, pr)))
     update_mock([:pr_commits], &(Map.put(&1, number, [])))
     update_mock([:comments], &(Map.put(&1, number, [])))
@@ -260,7 +261,8 @@ defmodule BorsNG.GitHub.FriendlyMock do
                      base_ref: base_ref,
                      head_sha: head_sha,
                      body: body,
-                     user: user}) do
+                     user: user,
+                     mergeable: mergeable}) do
     %{
     "number" => number,
     "title" => title,
@@ -281,6 +283,7 @@ defmodule BorsNG.GitHub.FriendlyMock do
     },
     "user" => user,
     "merged_at" => "some non-nul time :)",
+    "mergeable" => mergeable,
     }
   end
 
--- a/lib/github/github/pr.ex
+++ b/lib/github/github/pr.ex
@@ -12,6 +12,7 @@ defmodule BorsNG.GitHub.Pr do
     base_ref: bitstring,
     head_sha: bitstring,
     user: BorsNG.GitHub.User.t,
+    mergeable: boolean | nil,
   }
   defstruct(
     number: 0,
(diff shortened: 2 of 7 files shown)


noizwaves:
Hey there, I just wanted to give this PR a friendly little bump. If there is any other information or commentary that you'd like to help understand this PR, please don't hesitate to ask.

Thanks again! Adam

notriddle:
This is an amazingly good pull request.

bors r+

noizwaves:
@notriddle Thank you for the great feedback.

The DB mergeability flag is a great idea. We're definitely in favor of reducing API calls as well.

saifelse:
@noizwaves : Just sanity checking, before I investigate my situation further, is this feature still working the way you had intended? I'm in a situation where I have a PR that had a merge conflict with the base, but instead of failing fast, it ended up doing the bad bisect you outlined in [that forum post](https://forum.bors.tech/t/limiting-the-scope-of-merge-failures-in-batches/421).

One possibility is that GH returned `mergeable: nil` for the bad PR, but AFAICT, in my case the bad PR had a merge conflict for a while and the GH UI even showed it, so that case seems unlikely.

PR #862, opened 2020-01-31
Cut PR body in squash commits

indirect:
The bors.toml configuration option `cut_body_after` was being ignored if the option `use_squash_merge` was also set. This commit ensures that the squashed commit message will also respect the `cut_body_after` option rather than ignoring it.


indirect:
(I have tested this commit in production with both squash merges and clip body enabled, it works for me.)

PR #881, opened 2020-02-18
When retrying a merge-conflict branch, don't comment

### LINKED ISSUE #880: Bors makes repeatedly leaves the same comment on the issue
https://github.com/rust-analyzer/rust-analyzer/pull/3204#issuecomment-587442686

I think the story with this PR is as follows:

* PR is submited
* another PR is merged, which modifies `.gtihub/workflows`
* orignal PR is r+, but is not merged due to #806 
* I merge PR manually
* bors gets very talkative :D

notriddle:
In practice, on busy repositories, it's way too chatty.

Fixes #880

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -867,10 +867,14 @@ defmodule BorsNG.Worker.Batcher do
 
   defp send_message(repo_conn, patches, message) do
     body = Batcher.Message.generate_message(message)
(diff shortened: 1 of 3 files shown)


matklad:
Hm, I am not sure that the original bug was because of chattines: bors made a spree of comments with very short interval between them. So to me it looked like a bug with a retry logic which fired more times than it should have

PR #886, opened 2020-02-22
Improvements to prerun_timeout stuff

pksunkara:
Fixed the bug where bors is using the prerun_timout_sec not as the total time elapsed.

Also, removed the exponential backoff, since it get very big very soon. Changed the polling to a constant 5 mins.


notriddle:
`Function Github.map_changed_status/1 does not exist.`

I think it should be `GitHub.map_changed_status` probably?

pksunkara:
Oops. Fixed it 😄

pksunkara:
@notriddle I am not sure this actually fixed it. Check this PR. https://github.com/clap-rs/clap/pull/1706. I approved before all the checks passed. But it's still not merging. Do you know what happened here? Can we try to find the logs?

PR #888, opened 2020-02-24
Default bors deployments to use SSL

swist:
There is no reason really to not use SSL these days.
If someone wants to keep legacy behaviour this is still
available via an envvar


swist:
bors try

notriddle:
My initial thought was that I didn't want to merge this as-is, but since Heroku probably does the `X-Forwarded-Proto` thing, it should be fine.

bors r+

PR #928, opened 2020-04-10
Add Co-authored-by to squash commits.

gebner:
This PR adds the `Co-authored-by:` trailers to squash commits.  Non-squash commits already have these trailers since #392.


bryangingechen:
Should this (and #392) pick up co-authors from commits which are themselves co-authored?

notriddle:
@bryangingechen We could, but it doesn't seem necessary and I'm not going to block on it. Bors squashed commits use the message in the PR body, so if you want to add more people as co-authors, you're probably best off just including their co-author names in the PR body itself.

Issue #905, opened 2020-03-11
Bors run don't Batch PRs after a failure


ahlp:
Hello,
After update the bors, there is scenarios that it runs the PR without batching them.
All the runs don't have the `single` flag setted.
![image](https://user-images.githubusercontent.com/7283826/76445869-de162100-63a4-11ea-95df-79bd5e851749.png)

After fail the batch with 4 PRs, all PRs put in the queue after that are considered single runs.

Is this the correct behavior?
I was expecting it to continue to batch the new PRs.

longtimeago:
moreover, it looks like previously available approach (`bors r-`, `bors r+`) doesn't help either - PR is added in a separate run

unicolet:
I've spent a bit of time to familiarize with bors codebase and I'm beginning to think that the issue might have been caused by this line:

https://github.com/bors-ng/bors-ng/pull/839/files#diff-2969b4549f32c43c842acc6cdd64a94aL858

notice the PR removes the `limit(1)` clause, which means the query _might_ return more than one row. I tried adding a simple `Enum.take(1)` and it seems all tests still pass. I've opened https://github.com/bors-ng/bors-ng/pull/957 as a tentative fix

unicolet:
FYI we just deployed the change and so far bors appears to be batching as it should :crossed_fingers:

PR #957, opened 2020-05-17
#905 Always return at most one batch

unicolet:
after #839 (files) it _might_ have been possible for the query to retunr more than one row. This change effectively prevents that, by ensuring we always returning the first element.  In case no rows are returned by the query the call to `Enum.take` is safe, as `Enum.take([], 0)` simply returns an empty enumerable.

Fixes: #905


unicolet:
shoot I messed up, that should teach me to do thing hastily :-(

unicolet:
@notriddle gentle poke :-) thank you

notriddle:
Sure.

bors r+

PR #986, opened 2020-07-12
The required_approvals option should require up-to-date approvals

galbash:
### Overview
This is an attempt at implementing: https://forum.bors.tech/t/the-required-approvals-option-should-require-up-to-date-approvals/471

The implementation follows the RFC: a flag called `up_to_date_approvals` is added to `bors.toml`, which defaults to `false`. When this flag is set to `true` we validate that the reviews are associated with the current commit id that BORS is running on for the patch (and not just associated with the pr in general).

We still do the regular check for any PR reviews for two reasons:
1. We want to identify rejecting reviews, even if they are not associated with the last commit
2. We want to differentiate the error message of the case where a pr doesn't have enough reviewers and the case where the reviews are not up to date (since getting the original error message when you do have enough reviewers can be quite confusing, as Github's  association between a review and a commit isn't very clear from the UI)

### Current caveats of the implementation:
1. Unlike GitHub's behavior, when a user adds a merge commit that doesn't change the code to the repository, BORS will now require re-approves (where GitHub identifies this and don't mark the reviews as stale). The reason this is not implemented is that it would require a potentially large amount of API calls to GitHub (to fetch each commit's number of files changed) and since when using the BORS workflow you don't have to do merge commits, this seems fine. Would love to hear some input on how we can implement this in a way that is more cost-effective.
2. The status messages aren't as verbose as the RFC suggested (x out of Y are stale). The reason is that it will require much more code changes, given the current structure of the `GitHub` and `reviews` files. This again felt like it wasn't cost-effective enough.
3. Currently does not require up-to-date review from code owners (i.e. if 1 reviewer is required + code owner reviews are required, a PR will be accepted in the case where the code owner review is not up to date, but an additional up-to-date review exists which is not by a code owner). I am not sure if that is aligned with GitHub's behavior, but I think we can open a separate issue to incorporate this to code owner reviews if we choose to.

### Things to note:
1. The flag is currently off by default (same as GitHub's behavior for dismissing stale PRs) but this should probably be discussed.
2. Opening this early to get some guidance on what steps on the way I missed (opening an issue, some more RFC related stuff) and to let people know I am actively working on it. Also since this is my first time opening a PR for BORS, I am assuming that I overlooked some things so please be as scrutinous as possible :) 

### Open tasks:
Opened this ASAP to mark that I am working on it and prevent duplicate work & get fast feedback. Here is a list of things left to be done (should be updated over time).
- [x] test on a BORS instance (not just UT)

### DIFF
--- a/lib/github/github.ex
+++ b/lib/github/github.ex
@@ -263,6 +263,18 @@ defmodule BorsNG.GitHub do
     labels
   end
 
+  @spec get_commit_reviews!(tconn, integer | bitstring, binary) :: map
+  def get_commit_reviews!(repo_conn, issue_xref, sha) do
+    {:ok, labels} =
+      GenServer.call(
+        BorsNG.GitHub,
+        {:get_reviews, repo_conn, {issue_xref, sha}},
+        Confex.fetch_env!(:bors, :api_github_timeout)
+      )
+
+    labels
+  end
+
   @spec get_file!(tconn, binary, binary) :: binary | nil
   def get_file!(repo_conn, branch, path) do
     {:ok, file} =
--- a/lib/github/github/reviews.ex
+++ b/lib/github/github/reviews.ex
@@ -46,4 +46,15 @@ defmodule BorsNG.GitHub.Reviews do
     )
     |> Map.put("approvers", approved_by)
   end
+
+  @spec filter_sha!([map], nil) :: [map]
+  def filter_sha!(json, nil) do
+    json
+  end
+
+  @spec filter_sha!([map], binary) :: [map]
+  def filter_sha!(json, sha) do
+    json
+    |> Enum.filter(fn %{"commit_id" => commit_id} -> commit_id == sha end)
+  end
 end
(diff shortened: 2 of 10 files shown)


notriddle:
https://community-tc.services.mozilla.com/tasks/aFebAemVR_i-Kx2d4aYZhQ/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FaFebAemVR_i-Kx2d4aYZhQ%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log looks bad.

    lib/github/github/server.ex:437:no_return
    Function do_handle_call/4 has no local return.

I'm going to venture a guess and say that you need to add `:up_to_date_reviews` to the repo type definition.

galbash:
yep will look into that 👍  will also update here once I've tested it on a real bors instance (hopefully this weekend). Would you like me to convert it to a draft in the meantime?

notriddle:
I can just wait it in here. It's fine.

galbash:
Hi @notriddle, status checks are now passing + I tested it on a real BORS instance and it seems to be working fine. Added another caveat (#3) in the PR description (not requiring up to date review from code owners currently) - let me know if that's not good enough for now.
Also if there are any other code related review comments let me know :)

PR #1002, opened 2020-08-13
Don't simply ignore unset commit statuses

### LINKED ISSUE #1001: bug: bors doesn't handle unset commit statuses
[Here's](https://github.com/cockroachdb/cockroach/blob/master/.github/bors.toml#L3-L19) the bors config we use over at https://github.com/cockroachdb/cockroach:

```toml
# List of commit statuses that must pass on the merge commit before it is
# pushed to master.
status = ["Compile Build (Cockroach)"]

# List of commit statuses that must not be failing on the PR commit when it is
# r+-ed. If it's still in progress (for e.g. if CI is still running), bors will
# construct the merge commit in parallel and simply wait for success right
# before merging.
pr_status = ["license/cla", "GitHub CI (Cockroach)"]
```

We observe this bug in bors that I think is due to racey conditions around its reliance on webhooks. If you look at https://bors.crdb.io/repositories/56/log, you'll sometimes find the following:
image

The `bors r+` attempt in question there is https://github.com/cockroachdb/cockroach/pull/50446#issuecomment-673098315. If you look right above, the author amended their commit here https://github.com/cockroachdb/cockroach/pull/50446#event-3649360208 at Aug 12, 2020, 4:37 PM EDT. The author also almost instantly `bors r+`-ed their PR at Aug 12, 2020, 4:37 PM EDT. What I would expect to happen is that bors would wait for the "GitHub CI (Cockroach)" check on the amended commit to have run. What I think happened instead is that bors looked at the commit status for the commit before it was amended, and having seen that pass the CI check successfully, then went ahead and started the merge commit process. Unfortunately that's not behavior we want, I think what happens is that the as the merge commit checks complete and come back to find that the pr status has now failed (due to the amended commit), and thus crashing bors. This also has the unfortunate side effect of silently dropping the queue built up behind the PR in question, and all the other others who have `bors r+`-ed their PRs will not see any activity. That's what we see happen with this `bors r+` attempt: https://github.com/cockroachdb/cockroach/pull/52596#issuecomment-672021571.

irfansharif:
Fixes #1001.

Previously we treated unset statuses as :ok. This, I think, is
non-intuitive as it's entirely possible for CI to be slow to set even a
pending status. If someone `bors r+`s before that's done, bors would go
ahead and create the merge commit, which is not the contract defined by
the `pr_status` config (it's supposed to enumerate the list of commit statuses
that must pass on the PR commit when it is r+-ed).

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -916,29 +916,28 @@ defmodule BorsNG.Worker.Batcher do
       |> MapSet.new()
       |> MapSet.disjoint?(MapSet.new(toml.block_labels))
 
+    github_commit_statuses = GitHub.get_commit_status!(repo_conn, patch.commit)
+    pr_status_mapset = MapSet.new(toml.pr_status)
+
     no_error_status =
-      repo_conn
-      |> GitHub.get_commit_status!(patch.commit)
+      github_commit_statuses
       |> Enum.filter(fn {_, status} -> status == :error end)
       |> Enum.map(fn {context, _} -> context end)
       |> MapSet.new()
-      |> MapSet.disjoint?(MapSet.new(toml.pr_status))
+      |> MapSet.disjoint?(pr_status_mapset)
 
     no_waiting_status =
-      repo_conn
-      |> GitHub.get_commit_status!(patch.commit)
+      github_commit_statuses
       |> Enum.filter(fn {_, status} -> status == :running end)
       |> Enum.map(fn {context, _} -> context end)
       |> MapSet.new()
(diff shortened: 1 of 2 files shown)


irfansharif:
I don't actually know how to test this behavior, I couldn't get the repo running on my machine (though I could probably try a bit harder).

notriddle:
Tested at https://github.com/bors-ng-test-org/test-repo-2/pull/33 and found a bug. This seems to pretty much solve it.

irfansharif:
Thanks for the quick turnaround, @notriddle!

Issue #993, opened 2020-07-29
Bors won't merge with all codeowners approval


santiph:
After all codeowners approved it, BORS will still claim some are missing.
Source branch comes from a fork.

Our theory is that bors is ANDing all the possible owners/teams per file. While Github is ORing them.
Therefore bors requires brewery to approve everything when the *-Rule is there.
Which is no visible nor required by Github to its UI

CODEOWNERS file:
```
*                             @frontends/brewery
packages/events/**/*          @frontends/events
```

affected files on the PR:
```
packages/events/xing-event/Event/Event.tsx
packages/events/xing-event/Event/spec/__snapshots__/Event.spec.jsx.snap
packages/events/xing-event/package.json
packages/events/xing-event/styles/event.css
packages/events/xing-event/styles/ticket-price-range.css
packages/events/xing-event/styles/title.css
```

Bors logs: 
```
{:undef,
 [
   {nil, :id, [], []},
   {BorsNG.Worker.Batcher, :"-check_code_owner/3-fun-0-", 3,
    [file: 'lib/worker/batcher.ex', line: 1006]},
   {Enum, :any_list, 2, [file: 'lib/enum.ex', line: 2864]},
   {BorsNG.Worker.Batcher, :"-check_code_owner/3-fun-1-", 3,
    [file: 'lib/worker/batcher.ex', line: 1005]},
   {Enum, :any_list, 2, [file: 'lib/enum.ex', line: 2864]},
   {Enum, :"-map/2-lists^map/1-0-", 2,
    [file: 'lib/enum.ex', line: 1314]},
   {Enum, :"-map/2-lists^map/1-0-", 2,
    [file: 'lib/enum.ex', line: 1314]},
   {BorsNG.Worker.Batcher, :check_code_owner, 3,
    [file: 'lib/worker/batcher.ex', line: 987]}
 ]}
```

```
12:58:22.282 request_id=FiY6jqVPhw3-lx8AE2XB pid=<0.15008.7> [info] POST /webhook/github
12:58:22.284 request_id=FiY6jqVPhw3-lx8AE2XB pid=<0.15008.7> [info] Sent 200 in 1ms
12:58:33.257 pid=<0.15012.7> [info] CONNECTED TO BorsNG.UserSocket in 1ms
  Transport: :websocket
  Serializer: Phoenix.Socket.V2.JSONSerializer
  Parameters: %{"token" => "notyourtoken😅", "vsn" => "2.0.0"}
[...]

PR #1005, opened 2020-08-19
Unique list of reviewers for multiple code changes #933

ucwaldo:
This is a simple attempt to fix https://github.com/bors-ng/bors-ng/issues/993

While checking the `parse_owners.ex` file, i realized that it returns one reviewer per changed file even when all changes belong to the same codeowners.

To be honest, I did not dig very deep and it could be that this happens somewhere later in the pipeline already and i just missed it.

If that's the wrong approach because approvals are indeed tracked per file, i am happy to give it another go, ideally with some directions

### DIFF
--- a/lib/github/parse_owner.ex
+++ b/lib/github/parse_owner.ex
@@ -87,7 +87,10 @@ defmodule BorsNG.CodeOwnerParser do
         end)
       end)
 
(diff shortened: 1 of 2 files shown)


notriddle:
No, the GH Reviews interface is all built around the whole pull request being approved or disapproved.

ucwaldo:
Great, so my change should be valid. Thanks for the quick review @notriddle !

PR #1021, opened 2020-09-03
Add logging info and shorter poll interval to prerun

### LINKED ISSUE #1004: bug: bors does not automatically try to run when all required PR statuses are set
See https://github.com/cockroachdb/cockroach/pull/52991. I "bors r+"-ed the PR while the PR statuses were yet unset. Despite bors telling us that it would automatically retry, I don't think it did.

notriddle:
See https://github.com/bors-ng-test-org/test-repo-2/pull/34 for test run

Fixes #1004

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -39,7 +39,7 @@ defmodule BorsNG.Worker.Batcher do
 
   # Every half-hour
   @poll_period 30 * 60 * 1000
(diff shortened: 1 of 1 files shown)


irfansharif:
Do you happen to know why the 5m timeout that was there previously ended up dropping the request entirely, like the example linked in https://github.com/bors-ng/bors-ng/issues/1004#issue-681377505?

Issue #950, opened 2020-05-04
Using delete_merged_branches shouldn't delete if there are dependent branches


joshma:
I think there may be two issues here, but one is more important than the other:

1. Some engineers on the teams have dependent PRs out: master, feat-a, and feat-b, where feat-a is off of master and feat-b is off of feat-a. If feat-a is merged, delete_merged_branches=true causes bors to delete feat-a, which then causes an issue for feat-b. (GitHub closes the feat-b PR.)

2. Separately, we haven't actually started using bors to merge, and I think bors is deleting a PR branch even if bors isn't the one merging. In other words, we have the bors bot turned on and an engineer **could** say `bors r+` to merge it, but the engineer is instead using the UI merge button. In this scenario, bors is still deleting the branch after merging and thus breaking the dependent PR.

Fixing 1 would probably be sufficient - could bors leave a branch alone if there are PRs dependent on it?

Fixing 2 might be easier though - bors shouldn't delete branches that it hasn't merged itself.

browdus:
This is even more painful than it otherwise would be because it leaves the dependent PR (`feat-b` in the original example) in an unrecoverable state. The PR can't be opened because the base branch has been deleted, and the base branch can't be changed because the PR is closed.

berkowitzi:
Looks like this is solved by #1023 !

PR #1023, opened 2020-09-06
Change base before delete

champo:
A possible fix for one of the issues in https://github.com/bors-ng/bors-ng/issues/950. 
A proposal RFC for this is at https://forum.bors.tech/t/update-base-branch-of-prs-depending-on-a-branch-that-will-be-deleted/507

This PR adds an opt-in feature where if bors is going to delete a branch (due to a PR being merged), it will first update the base branch of other PRs that depend on it. It does so simply by changing the `base_ref` using the github API. This won't cause changes to the commit history, it will simply keep the dependant PRs open (otherwise they are automatically closed) and showing the new commits only.

### DIFF
--- a/lib/github/github.ex
+++ b/lib/github/github.ex
@@ -103,6 +103,18 @@ defmodule BorsNG.GitHub do
     prs
   end
 
+  @spec get_open_prs_with_base!(tconn, binary) :: [tpr]
+  def get_open_prs_with_base!(repo_conn, base) do
+    {:ok, prs} =
+      GenServer.call(
+        BorsNG.GitHub,
+        {:get_open_prs_with_base, repo_conn, {base}},
+        100_000
+      )
+
+    prs
+  end
+
   @spec push!(tconn, binary, binary) :: binary
   def push!(repo_conn, sha, to) do
     {:ok, sha} =
--- a/lib/github/github/server.ex
+++ b/lib/github/github/server.ex
@@ -172,6 +172,15 @@ defmodule BorsNG.GitHub.Server do
      )}
   end
 
+  def do_handle_call(:get_open_prs_with_base, {{:raw, token}, repo_xref}, {base}) do
+    {:ok,
+     get_open_prs_!(
+       token,
+       "#{site()}/repositories/#{repo_xref}/pulls?state=open&base=#{base}",
+       []
+     )}
+  end
+
   def do_handle_call(:push, repo_conn, {sha, to}) do
     repo_conn
     |> patch!("git/refs/heads/#{to}", Poison.encode!(%{sha: sha}))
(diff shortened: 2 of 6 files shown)


notriddle:
It’s acceptable, though I’ll want to open an RFC on the forum before merging it.

I’m curious, though, why you don’t just not delete the branch when there are PRs referencing it? That’s what most of the comments on the issue seem to want, and it seems simpler than this.

champo:
> I’m curious, though, why you don’t just not delete the branch when there are PRs referencing it? That’s what most of the comments on the issue seem to want, and it seems simpler than this.

Not deleting the branch feels like solving it mid way to me. Let me explain why. Say you have 2 PRs A and B, where B points to A's base branch `branch_a` and A to the `default` branch. After `A` is merged, if `B` is merged it's probably going to be a mistake: the branch it's now been integrated to is not included in any PR and is actually considered merged. Feels like a recipe for easy mistakes. Except for a very specific workflow, I'd expect you want `B` to merge onto the `default` branch.

> It’s acceptable, though I’ll want to open an RFC on the forum before merging it.

Fair enough, I'll look into the RFC process! Thanks for the quick reply :)

notriddle (approved):
bors r+

nomeata:
This doesn't fully work as desired together with `use_squash_merge`, as GitHub won't realize that half the commits are now in the base branch, right?

To work with squashed commits, bors would have to push a commit to the feature branch that merges the new base branch into it; possibly like describe at https://www.joachim-breitner.de/blog/767-git_post-squash.

Did you consider that?

notriddle:
I never really considered that, no. It might help, and I would accept an RFC + Pull Request to implement it, but I personally don’t use squash commits, because of bugs like this one, especially the one with marking PRs as Merged.

The proposed workaround also seems pretty complicated, because it has bors writing to the user’s feature branches, and also has the user writing to the user’s feature branches. This is [shared mutable state](https://manishearth.github.io/blog/2015/05/17/the-problem-with-shared-mutability/), with all the pain points and potential for silent data loss. The data is actually still in your reflog, at least until it gets GC-ed, but how long will it take for you to even notice?

Also it could be a bit annoying, since you need to run `git pull --rebase` whenever bors pushes to the branch.

I’m not actually against adding this as an optional feature, but I’m not volunteering to implement it, because I wouldn’t want to use it.

nomeata:
Yes, I agree it is far from great. It may be better than leaving the users alone in the woods when it comes to stacked PRs (assuming you want the pretty `master` history you get from squash merges).

I don’t think data loss is a problem – all commits on feature branches, whether from bors or from users, can (and should) be non-force pushes in this mode. But there are certainly pain points – feature branches will look like they contain all the commits from the already merged dependent branches (but the diff will be clean), for example. And you need to `git pull --no-ff` indeed.

And it gets even more complicated (but still possible) when one doesn’t require strict merges, and the feature branch ends up having a different branch point than the merged commit, and may end up being too complicated for the git-non-nerds…

Anyways, I was just wondering if the bors community might have a better answer…

PR #1032, opened 2020-09-16
Fix codeowners with double asterisk and prefixed with /

dieend:
If the pattern contains a double-asterisk and prefixed with '/', the code failed to parse it correctly.

### DIFF
--- a/lib/github/parse_owner.ex
+++ b/lib/github/parse_owner.ex
@@ -112,8 +112,16 @@ defmodule BorsNG.CodeOwnerParser do
 
       String.contains?(file_pattern, double_asterisk) ->
         patterns = String.split(file_pattern, double_asterisk, parts: 2)
+        parent = List.first(patterns)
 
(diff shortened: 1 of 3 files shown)


lib/github/parse_owner.ex
```diff
@@ -112,8 +112,16 @@ defmodule BorsNG.CodeOwnerParser do
...
       String.contains?(file_pattern, double_asterisk) ->
         patterns = String.split(file_pattern, double_asterisk, parts: 2)
+        parent = List.first(patterns)
 
-        String.starts_with?(file_name, List.first(patterns)) &&
+        first_part =
+          if String.starts_with?(parent, "/") do
+            String.starts_with?("/" <> file_name, parent)
+          else
+            String.starts_with?(file_name, List.first(patterns))
```
notriddle, line 121:
```suggestion
            String.starts_with?(file_name, parent)
```

notriddle: Also, does something similar need to be done for the other two cases?

PR #1082, opened 2020-11-16
Log unknown GitHub events in webhook

zli-simspace:
While troubleshooting an issue, I see in our logs.

```
    ** (FunctionClauseError) no function clause matching in BorsNG.WebhookController.do_webhook/3
```

Unfortunately the error that follows shows only the first argument `conn` truncated and not the important second and third arguments.

## Solution

Similar to `do_webhook_pr`, I've added a `do_webhook` clause to log the unknown cases. I have it return a 404 (so its close to the previous behaviour of returning a 500) but suggestions for a different behaviour are welcome.

## Test

Tested by running

```
curl -X POST localhost:8000/webhook/github -H 'x-github-event: bogus' -d "{}"
```

with signature checking temporarily disabled and saw

```
[info] WebhookController: Got unknown Github event: bogus
```

in the server output.

### DIFF
--- a/lib/web/controllers/webhook_controller.ex
+++ b/lib/web/controllers/webhook_controller.ex
@@ -91,7 +91,7 @@ defmodule BorsNG.WebhookController do
   end
 
   # Deprecated event, but still being sent since 2020.
-  def do_webhook(conn, "github", "integration_installation") do
+  def do_webhook(_conn, "github", "integration_installation") do
     :ok
   end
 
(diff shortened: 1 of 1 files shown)


notriddle (approved):

notriddle:
```
** (Mix) mix format failed due to --check-formatted.

The following files were not formatted:

  * lib/web/controllers/webhook_controller.ex
```

zli-simspace:
Thanks! Should be fixed now.

Reran `mix format --check-formatted` locally to confirm.

notriddle (approved):
bors r+

PR #1083, opened 2020-11-16
Fix MatchError from wrong number of values returned.

zli-simspace:
In commit 9c555f8cec72fe1891429534d5dd81d7c436fc18, the result of the first branch in

```
      case toml_result do
        {:ok, toml} -> {toml.delete_merged_branches, toml.update_base_for_deletes}
        _ -> false
      end
```

is updated but not the second branch, resulting in a MatchError.

This small PR fixes that.


notriddle (approved):
bors r+

PR #1116, opened 2020-12-31
Fix crash on self_check because of trying to start multiple process

### LINKED ISSUE #1085: A failing check_self can kill the GenServer
I'm seeing periodic restart in our (private) Bors server. A number of them are preceded by a failed `self_check``

```
[error] GenServer #PID<0.20129.2> terminating
** (MatchError) no match of right hand side value: #PID<0.20128.2>
    (bors) lib/worker/batcher.ex:1206: BorsNG.Worker.Batcher.check_self/1
    (bors) lib/worker/batcher.ex:231: BorsNG.Worker.Batcher.handle_info/2
    (stdlib) gen_server.erl:637: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:711: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: {:poll, :repeat}
[info] Running BorsNG.Endpoint with cowboy <version> at <ip>
```

I know the self-check error itself is normal but from the timestamps, it does seem to be the trigger for the Bors restart. It could still be a number of other things though.

Sometimes, before the last line indicating a restart, there's also

```
pid=<0.8508.0> [error] GenServer BorsNG.Worker.Batcher.Registry terminating
** (stop) exited in: GenServer.call(BorsNG.Worker.Batcher.Supervisor, {:start_child, {{BorsNG.Worker.Batcher, :start_link, 'l'}, :permanent, 5000, :worker, [BorsNG.Worker.Batcher]}}, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir) lib/gen_server.ex:924: GenServer.call/3
    (bors) lib/worker/batcher/registry.ex:50: BorsNG.Worker.Batcher.Registry.do_start/1
    (bors) lib/worker/batcher/registry.ex:55: BorsNG.Worker.Batcher.Registry.start_and_insert/2
    (bors) lib/worker/batcher/registry.ex:83: BorsNG.Worker.Batcher.Registry.handle_info/2
    (stdlib) gen_server.erl:637: :gen_server.try_dispatch/4
    (stdlib) gen_server.erl:711: :gen_server.handle_msg/6
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Last message: {:DOWN, #Reference<0.1078022152.3600285698.175770>, :process, #PID<0.14783.0>, :shutdown}
```

I came across this as I'm still looking into why sometimes preflight polling stops for no apparent reason. And I think this is indeed the cause. (One way to help with that is to persist the preflight poll into the database. I kind of regret not having spent my time implementing that instead of trying to track this bug.)

I don't yet know why a failed self_check would end the main/initial GenServer.

I'm also trying to see if others are seeing this as well or not (grep for `Running BorsNG.Endpoint with cowboy` in the logs).

dieend:
This change will remove `check_self` crash.

It hasn't been tested outside of the private fork.

Fixing this because unable to resolve #1085.

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -78,6 +78,7 @@ defmodule BorsNG.Worker.Batcher do
   # Server callbacks
 
   def init(project_id) do
+    BorsNG.Worker.Batcher.Registry.monitor(self(), project_id)
     Process.send_after(
       self(),
       {:poll, :repeat},
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -79,6 +79,7 @@ defmodule BorsNG.Worker.Batcher do
 
   def init(project_id) do
     BorsNG.Worker.Batcher.Registry.monitor(self(), project_id)
+
     Process.send_after(
       self(),
       {:poll, :repeat},
--- a/lib/worker/batcher/registry.ex
+++ b/lib/worker/batcher/registry.ex
@@ -18,6 +18,8 @@ defmodule BorsNG.Worker.Batcher.Registry do
   alias BorsNG.Database.Project
   alias BorsNG.Database.Repo
 
+  require Logger
+
   @name BorsNG.Worker.Batcher.Registry
 
   # Public API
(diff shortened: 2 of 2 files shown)


notriddle:
~~~
lib/worker/batcher/registry.ex:98:call_to_missing
Call to missing or private function Logger.warning/1
~~~

If it fixes this problem, and you fix the CI failure, then I’ll happily merge it.

notriddle (approved):
bors r+

PR #1128, opened 2021-01-15
Fix CODEOWNERS failure due to hardcoded "master"

quicksnap:
Our repository does not use `master` as the mainline for bors. Bors is erroneously passing the codeowners check, and in the logs it would get a 404: 
```
[error] GET https://api.github.com/repositories/123456/contents/.github/CODEOWNERS -> 404 (127.281 ms)
```

**I have not tested this PR**. I do not know Elixir, but this seems to be a good starting point for a change. An improvement could be that if `use_codeowners` is true and the http request fails, the codeowners check should fail.

Hopefully this is a good jumping off point. My coworker knows Elixir and has contributed in the past, so maybe they can get this PR into a better state if it's not acceptable as-is :)

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -992,7 +992,7 @@ defmodule BorsNG.Worker.Batcher do
       true
     else
       Logger.info("Checking code owners")
-      {:ok, code_owner} = Batcher.GetCodeOwners.get(repo_conn, "master")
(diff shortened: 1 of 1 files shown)


quicksnap:
Seems like this change has raised some CI hackles ;)

notriddle (approved):
bors r+

notriddle:
Not hard to fix the problem, now that you've pointed it out. Thanks!

PR #1131, opened 2021-01-17
Handle approvals of individual code owners

### LINKED ISSUE #1127: Bors rejects approved PRs that touch files owned only by individuals
In my project some files are owned by teams and some by individuals. Sometimes fully approved PRs are still rejected by bors with a:

> 👎 Rejected because of missing code owner approval

It turns out these PRs touched files owned **only** by individuals. Bors seems to [only take team ownership into account](https://github.com/bors-ng/bors-ng/blob/5f899a527ea69feda2ebf9c7442281245038b0a6/lib/worker/batcher.ex#L1009:L1033) thereby rejecting individually owned files by default even though their owners approved them.

It seems that `if-then` expression needs to contain an `else` branch for individual owners like: 

```
  ...
else
  Enum.any?(passed_review["approvers"], fn username -> 
    required == username
  end)
end
```

basvandijk:
This hopefully fixes #1127. This has not been tested.

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -1028,6 +1028,10 @@ defmodule BorsNG.Worker.Batcher do
 
               Logger.info("Approved: #{inspect(team_approved)}")
               team_approved
+            else
+              Enum.any?(passed_review["approvers"], fn username ->
+                required == username
+              end)
             end
           end)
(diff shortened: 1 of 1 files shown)


notriddle (approved):
bors r+

PR #1134, opened 2021-01-23
Make the database timeout configurable

asbach:
Add the ability to configure the database timeout using an environment variable.
The default value remains at 15 seconds, but for repositories with 1000s of users, making the value configurable avoids running into issues.

#1133 already helped, but we have still seen some timeouts adding users.


notriddle:
Opened up a thread on the forum for it.

https://forum.bors.tech/t/make-the-database-timeout-configurable/547

PR #1142, opened 2021-02-01
Remove '@' from the code owner before comparing it to the approver

### LINKED ISSUE #1127: Bors rejects approved PRs that touch files owned only by individuals

basvandijk:
This should really fix #1127

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -1031,7 +1031,7 @@ defmodule BorsNG.Worker.Batcher do
               team_approved
             else
               Enum.any?(passed_review["approvers"], fn username ->
-                required == username
+                String.slice(required, 1, String.length(required) - 1) == username
               end)
             end
           end)


basvandijk:
@notriddle it turns out my fix in https://github.com/bors-ng/bors-ng/pull/1131 was incomplete. I also have to remove the "@" prefix from code owners before comparing them to the approvers. It would be great if you could review and merge this.

notriddle (approved):
bors r+

basvandijk:
Thanks for the merge! I can confirm this actually fixes the problem. Bors now correctly handles individual code owners.

PR #1155, opened 2021-02-14
Support all of the CODEOWNERS locations documented by GitHub

### LINKED ISSUE #803: Check all CODEOWNERS locations
According to the GitHub documentation, codeowners can be defined at three locations.

> To use a CODEOWNERS file, create a new file called CODEOWNERS in the root, docs/, or.github/ directory of the repository, in the branch where you'd like to add the code owners.

bors only checks `.github/CODEOWNERS`  ([code](https://github.com/bors-ng/bors-ng/blob/fcec51b5f6e5a691d3e310bd35ba7b74e55b667f/lib/worker/batcher/get_code_owners.ex#L12)).

I believe that bors should check all three locations for the presence of a CODEOWNERS file.

I also think that it would be a good idea to crash/abort if `use_codeowners` is enabled, but no codeowners file could be found.

BuckyMaler:
As documented [here](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-file-location), GitHub supports `CODEOWNERS` in the root, `docs/`, or `.github/` directory of a repo.

Fixes #803

### DIFF
--- a/lib/worker/batcher/get_code_owners.ex
+++ b/lib/worker/batcher/get_code_owners.ex
@@ -9,7 +9,10 @@ defmodule BorsNG.Worker.Batcher.GetCodeOwners do
 
   @spec get(GitHub.tconn(), binary) :: {:ok, BorsNG.CodeOwners.t()} | {:error, terror}
   def get(repo_conn, branch) do
-    file = GitHub.get_file!(repo_conn, branch, ".github/CODEOWNERS")
+    file =
+      Enum.find_value(["CODEOWNERS", "docs/CODEOWNERS", ".github/CODEOWNERS"], fn path ->
+        GitHub.get_file!(repo_conn, branch, path)
+      end)
 
     BorsNG.CodeOwnerParser.parse_file(file)
   end
--- a/test/batcher/batcher_test.exs
+++ b/test/batcher/batcher_test.exs
@@ -605,6 +605,216 @@ defmodule BorsNG.Worker.BatcherTest do
     assert [] == Repo.all(Batch)
   end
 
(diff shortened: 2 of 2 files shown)


BuckyMaler:
This PR is the extent of my elixir experience. If I understand correctly, with these changes users that store `CODEOWNERS` in `docs/` will make an additional network request, which will result in degraded performance. That goes double for users that store `CODEOWNERS` in `.github/`. The degraded performance should be small though.

BuckyMaler:
Supporting a string value for `use_codeowners`, where the string is the path to the `CODEOWNERS` file, would eliminate the concerns I mentioned above. I think that may be the way to go. What do you think?

notriddle:
bors r+

> If I understand correctly, with these changes users that store CODEOWNERS in docs/ will make an additional network request, which will result in degraded performance.

I'm fine with it as is. The docs say that our CODEOWNERS work exactly the same as GitHub's, so our CODEOWNERS should work exactly like their does and any deviation is a bug.

> Supporting a string value for use_codeowners, where the string is the path to the CODEOWNERS file, would eliminate the concerns I mentioned above. I think that may be the way to go. What do you think?

Why not both?

notriddle:
bors r-

But first, I need to actually test this.

BuckyMaler:
> Why not both?

Do you mean support `use_codeowners` as a boolean or string? That's what I meant. I can open a separate PR for that. To prevent a breaking change `true` can be treated equivalently to `".github/CODEOWNERS"`.

notriddle (approved):
bors r+

PR #1171, opened 2021-02-25
Get all statuses from the pagination

### LINKED ISSUE #1169: Follow pagination for statuses
The call to get the statuses for a commit should follow the pagination as in https://docs.github.com/en/rest/reference/repos#list-commit-statuses-for-a-reference 

Currently if a repository has too many statuses, bors ends up missing some and rejecting PRs for the `pr_status` rule.

dpp23:
Get all the pages from the status api both for statuses and checks.

An alternative approach will be to simply bump the page size to 100 which should be good enough for most projects. The current limit of 30 is however not enough.

Fixes #1169

### DIFF
--- a/lib/github/github/server.ex
+++ b/lib/github/github/server.ex
@@ -398,47 +398,19 @@ defmodule BorsNG.GitHub.Server do
     end
   end
 
-  def do_handle_call(:get_commit_status, repo_conn, {sha}) do
+  def do_handle_call(:get_commit_status, {{:raw, token}, repo_xref}, {sha}) do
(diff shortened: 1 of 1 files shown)


notriddle:
https://github.com/bors-ng/bors-ng/runs/1983668763

dpp23:
> https://github.com/bors-ng/bors-ng/runs/1983668763

@notriddle sorry - wasn't aware of the formatter. Fixed now.

notriddle:
No problem. I just switched CI provider, which left this a bit broken.

bors r+

PR #1185, opened 2021-03-02
Update to newer versions of ecto and phoenix

notriddle:
The newest version of Ecto doesn't seem to be working on OTP 22.

I can live with that, but we need to wait until OTP 24 is released first.


notriddle (commented):
bors r+

PR #1191, opened 2021-03-16
Add cancel as an alias for bors r-

### LINKED ISSUE #1170: Make "bors cancel" an alias for "bors r-"
Given that "bors merge" was introduced as an alias for "bors r+", I propose making "bors cancel" be an alias for "bors r-". I've seen a lot of developers who has been confused about what "bors r+" really means, so I think "bors cancel" also will improve usability a little.

Also, don't forget to update [the documentation](https://bors.tech/documentation/) if this change is accepted.

tharun208:
This pr adds the alias `cancel` for `r-`

Fixes #1170

### DIFF
--- a/lib/web/command.ex
+++ b/lib/web/command.ex
@@ -162,6 +162,7 @@ defmodule BorsNG.Command do
   def parse_cmd("ping" <> _), do: [:ping]
   def parse_cmd("p=" <> rest), do: parse_priority(rest)
   def parse_cmd("retry" <> _), do: [:retry]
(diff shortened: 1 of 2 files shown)


tharun208:
@notriddle can we merge this?

notriddle:
Since this is a user-visible change, I opened a thread [on the forum](https://forum.bors.tech/t/add-bors-cancel-as-alias-for-bors-r/561). It's pretty uncontroversial, so I'll probably merge it in a few weeks with no problem, but I like for people to know about new features before they're merged, because once we add a new feature, it's a lot harder to remove it later.

PR #1195, opened 2021-03-22
Create health check route

### LINKED ISSUE #1192: Unable to utilize GCP Health Checking
Hello, I've been working on a Helm/K8S deployment for a private instance of Bors-ng and have been struggling to get load balancing working due to the fact that all the routes I've tried to health check on immediately redirect. Specifically, I'm deploying bors-ng to a Kubernetes cluster on Google Cloud Platform and their ingress controller/load balancer model automatically creates health checks that are required to pass before sending traffic to the backend service, in this case Bors. They require the health check endpoints to return 200 status codes unfortunately. I've looked into disabling the health checking or bypassing it somehow on GCP, but was unsuccessful in that approach and would prefer to have the health checks regardless.

Is there an existing route that I can use for health checking, or should I implement a simple endpoint for this case? I'm happy to do the implementation myself and make a PR for it.

RetWolf:
This PR implements the RFC documented here: https://forum.bors.tech/t/add-a-simple-health-check-endpoint/563 by creating a new `ServerController` meant to handle server specific actions such as health checking. This controller could be expanded upon later to have more routes related to the internals of the Elixir server.

Adds an endpoint at `/health` which will always return a HTTP 200 OK status code as long as the Elixir server is running

Fixes #1192


notriddle:
Sure.

bors r+

PR #1209, opened 2021-03-31
Helm chart to deploy Bors and Terraform examples 

carlosjgp:
I have created a helm chart to deploy `bors` internally on my company

I'd happily add some documentation to the readme file about this and the Terraform examples that I've also used internally

There is another Github action that could be added to host `bors` chart using Github pages
- https://github.com/helm/chart-releaser
- https://github.com/helm/chart-releaser-action
but it requires some configuration from your side to provide the necessary access/secrets
I'd create a new PR after this one is merged for that :D

After publishing the helm chart we could also add it to https://artifacthub.io/ by login and creating an organisation or just adding a repository

### DIFF
--- a/.github/workflows/config/chart-lint.yaml
+++ b/.github/workflows/config/chart-lint.yaml
@@ -1,0 +1,5 @@
+chart-dirs:
+  - deploy/charts
+
+chart-repos:
+  - bitnami=https://charts.bitnami.com/bitnami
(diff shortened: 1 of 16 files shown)


carlosjgp:
> I’m certainly not opposed to it, but since a Helm chart basically exposes an API and needs to be kept stable, this should definitely go through an RFC. Those are posted https://forum.bors.tech/c/development/draft-rfcs/11

https://forum.bors.tech/t/helm-chart-and-examples-using-terraform-with-kubernetes/566

PR #1221, opened 2021-04-12
Escape `@mentions` in PR body

### LINKED ISSUE #984: When adding PR description to the merge comment, escape @mentions
Github sends a notification if you are @mentioned in a commit message. This is unfortunate, because you'll get repeated notifications when the commit is rebased (or when you try several merges). Arguably, this is a bug in GitHub, but it seems like it's easier to fix this on bors' side, by escaping all `@mention`s

notriddle:
Fixes #984


DJMcNab:
What can we do to push this forward? `bevy` has run into the underlying issue a couple of times.

notriddle:
This should fix the problem. I made a mistake in the test code.

bors r+

PR #1223, opened 2021-04-14
Allow poll_period to be configured as an env variable

### LINKED ISSUE #1143: Adjustable poll_period in BorsNG.Worker.Batcher
We run our own instance of Bors-NG (thanks!) as part of our GitOps pipeline to make sure that configs are valid before trying to deploy them. The tests are designed to run quickly (60-90s tops). Sometimes the Github webhook is missed[1], it can take a long time for Bors-NG to realize that the merge can be completed. It looks like Bors-NG has a hard-coded 30 minute backup poll here:
https://github.com/bors-ng/bors-ng/blob/00a9c17725a91ec6a22f3538c2183d43fb690200/lib/worker/batcher.ex#L41

Would it be possible to make such a value configurable for those that run the infra? Or maybe a setting that could be set for the repo? I'm not all that picky, just looking for a way to reduce the value without having to fork/build our own container - so I'm open to any suggestions.

[1] Likely related to our infra we are working to fix.

dmorina:
closes #1143


notriddle (approved):
bors r+

Issue #1263, opened 2021-06-08
bors is down?


haampie:
https://app.bors.tech/

> Application error
> 
> An error occurred in the application and your page could not be served. If you are the application owner, check your logs for details. You can do this from the Heroku CLI with the command
heroku logs --tail

hi-rustin:
cc: @notriddle 

Could you please take a look? Is it related to my changes?

hi-rustin:
It should be the result of my changes, sorry!

haampie:
thank you!

PR #1264, opened 2021-06-08
Correct the base URL

hi-rustin:
close https://github.com/bors-ng/bors-ng/issues/1263


hi-rustin:
cc: @notriddle Could you please take a look?

Shouldn't the host be used here, but the full URL?

notriddle:
Yeah, it's probably the result of this.

Issue #1152, opened 2021-02-11
bors-ng silently discards all batches on GitHub 422 errors


asbach:
We have observed the following bors-ng behavior running on our own bors instance:

Crash log in bors:
```
{{:badmatch,
  {:error, :push, 422,
   "{\"message\":\"Required status check \\\"continuous-integration/jenkins/pr-merge\\\" is expected. At least 1 approving review is required by reviewers with write access.\",\"documentation_url\":\"https://docs.github.com/enterprise/2.22/user/articles/about-protected-branches\"}"}},
 [
   {BorsNG.Worker.Batcher, :complete_batch, 3,
    [file: 'lib/worker/batcher.ex', line: 712]},
   {BorsNG.Worker.Batcher, :maybe_complete_batch, 1,
    [file: 'lib/worker/batcher.ex', line: 658]},
   {BorsNG.Worker.Batcher, :handle_cast, 2,
    [file: 'lib/worker/batcher.ex', line: 94]},
   {:gen_server, :try_dispatch, 4,
    [file: 'gen_server.erl', line: 637]},
   {:gen_server, :handle_msg, 6,
    [file: 'gen_server.erl', line: 711]},
   {:proc_lib, :init_p_do_apply, 3,
    [file: 'proc_lib.erl', line: 249]}
 ]}
```

In this particular case, the target branch is has the required reviews set, but this particular branch shouldn't be used with bors anways, so that part is not what I'm worried about, I would say working as intended.

The issue that is actually a problem is that bors will silently drop all in-flight batches when it encounters this error. We have not yet taken a closer look into why this happens, but I have seen similar reports in other issues, e.g. #703. The other issues only mention that one batch and not all are dropped. 
Any ideas where we could start taking a closer look into the code?

tommilligan:
Hey @asbach, I believe this can be now closed as fixed by #1273. Please take a look and verify if it covers your use case, or provide more details if not.

asbach:
Hello @tommilligan,
I've deployed a version of bors that includes your fix and the issue appears to be fixed. Thanks a lot :-)

Issue #1150, opened 2021-02-10
Occasional Crash over “Required status check”


tiagolobocastro:
Hi,

We use bors.tech with our github and occasionally we see a crash:
```
{{:badmatch,
  {:error, :push, 422,
   "{\"message\":\"Required status check \\\"commitlint\\\" is expected.\",\"documentation_url\":\"https://docs.github.com/articles/about-protected-branches\"}"}},
 [
   {BorsNG.Worker.Batcher, :complete_batch, 3,
    [file: 'lib/worker/batcher.ex', line: 712]},
   {BorsNG.Worker.Batcher, :maybe_complete_batch, 1,
    [file: 'lib/worker/batcher.ex', line: 658]},
   {BorsNG.Worker.Batcher, :handle_cast, 2,
    [file: 'lib/worker/batcher.ex', line: 94]},
   {:gen_server, :try_dispatch, 4,
    [file: 'gen_server.erl', line: 637]},
   {:gen_server, :handle_msg, 6,
    [file: 'gen_server.erl', line: 711]},
   {:proc_lib, :init_p_do_apply, 3,
    [file: 'proc_lib.erl', line: 249]}
 ]}
```

That `commmitlint` is indeed a required status, though not sure why it's causing the crash.
Not sure if related to this or not but we've also observed that sometimes the PR's get stuck and we have to rebase, force push and retry, though since the crash does not mention a specific build/pr not sure how to correlate those things.

Any clues what could be the issue here?
Thank you!

jkryl:
I have the same problem (on the same project). It happened in the second run of tests when bors was trying to bisect commits. The second run on subset of PRs passed and therefore the PR should have been merged but bors crashes instead. Of course the commitlint check passed so even the message about required status check is wrong.

tommilligan:
Hey @tiagolobocastro , I believe this can be now closed as fixed by #1273. Please take a look and verify if it covers your use case, or provide more details if not.

The crash is caused when bors tries to merge a commit into your master branch, on which there is branch protection setup (and not fulfilled). My guess is that the `commitlint` check is still running on the new commit that bors created out of a subset of commits, at the time bors tries to merge the commit.

#1273 probably won't get the commit merging in in this case, but at least will provide a clear error message and not crash.

tiagolobocastro:
Brilliant thanks @tommilligan :+1: I'll close this one then. (can always reopen if it happens again)

Issue #1027, opened 2020-09-09
Handle HTTP 422 errors from GitHub


rvl:
Hi, we are getting "Cancelled" status and crashes when trying to merge a PR. It's a simple PR - single file changed, commit is GPG signed, no squash merges - nothing special.

What has happened is that bors-ng accepted the PR and attempted to merge, but GitHub rejected it due to absence of CODEOWNERS approvals.

I see multiple reports of this issue: #561 #674 #993 #725 #754. And I think @umamaistempo even fixed it once. So this might be a regression.

### Expected behaviour

Bors sends a comment to the PR saying that github rejected the merge with error code 422.

### Actual behaviour

`badmatch` crash dump is only visible in bors-ng history dashboard.

### Example PR

input-output-hk/cardano-node#1823

### Version

a4d04fd740e24d03833cd46a76f0f9e5be96f818 docker image running on our own instance https://bors-ng.aws.iohkdev.io/repositories/13/log#batch-6434

umamaistempo:
:thinking: 

Could you provide the stacktrace for that failure ?

rvl:
Sure... 9/10/2020, 7:08:33 AM | Crash | batch

```
{{:badmatch,
  {:error, :push, 422,
   "{\"message\":\"Waiting on code owner review from dcoutts, erikd, Jimbo4350, mrBliss, and/or newhoggy.\",\"documentation_url\":\"https://docs.github.com/articles/about-protected-branches\"}"}},
 [
   {BorsNG.Worker.Batcher, :complete_batch, 3,
    [file: 'lib/worker/batcher.ex', line: 707]},
   {BorsNG.Worker.Batcher, :maybe_complete_batch, 1,
    [file: 'lib/worker/batcher.ex', line: 653]},
   {BorsNG.Worker.Batcher, :handle_cast, 2,
    [file: 'lib/worker/batcher.ex', line: 92]},
   {:gen_server, :try_dispatch, 4,
    [file: 'gen_server.erl', line: 637]},
   {:gen_server, :handle_msg, 6,
    [file: 'gen_server.erl', line: 711]},
   {:proc_lib, :init_p_do_apply, 3,
    [file: 'proc_lib.erl', line: 249]}
 ]}
```

notriddle:
Do you have the `required_approvals` setting turned on in your bors.toml file?

rvl:
[...]

Issue #432, opened 2018-06-20
Crash merging PR


palango:
We ran into the following crash. I think it hapened after the build succeded and bors tried to merge into master.

```
{{:badmatch, {:error, :push}},
 [{BorsNG.GitHub, :push!, 3,
   [file: 'lib/github/github.ex', line: 63]},
  {BorsNG.Worker.Batcher, :complete_batch, 3,
   [file: 'lib/worker/batcher.ex', line: 381]},
  {BorsNG.Worker.Batcher, :maybe_complete_batch, 1,
   [file: 'lib/worker/batcher.ex', line: 368]},
  {BorsNG.Worker.Batcher, :handle_cast, 2,
   [file: 'lib/worker/batcher.ex', line: 77]},
  {:gen_server, :try_dispatch, 4,
   [file: 'gen_server.erl', line: 616]},
  {:gen_server, :handle_msg, 6,
   [file: 'gen_server.erl', line: 686]},
  {:proc_lib, :init_p_do_apply, 3,
   [file: 'proc_lib.erl', line: 247]}]}
```

JensRantil:
Looks like there's been a few [similar issue submissions](https://github.com/bors-ng/bors-ng/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+%3Apush%21). Could this be a duplicate of any of those?

vlad20012:
I've get something similar [here]({{:badmatch, {:error, :get_branch}},
 [
   {BorsNG.GitHub, :get_branch!, 2,
    [file: 'lib/github/github.ex', line: 70]},
   {BorsNG.Worker.Batcher, :start_waiting_batch, 1,
    [file: 'lib/worker/batcher.ex', line: 247]},
   {BorsNG.Worker.Batcher, :poll_, 1,
    [file: 'lib/worker/batcher.ex', line: 199]},
   {BorsNG.Worker.Batcher, :handle_info, 2,
    [file: 'lib/worker/batcher.ex', line: 181]},
   {:gen_server, :try_dispatch, 4,
    [file: 'gen_server.erl', line: 637]},
   {:gen_server, :handle_msg, 6,
    [file: 'gen_server.erl', line: 711]},
   {:proc_lib, :init_p_do_apply, 3,
    [file: 'proc_lib.erl', line: 249]}
 ]})
```
{{:badmatch, {:error, :get_branch}},
 [
   {BorsNG.GitHub, :get_branch!, 2,
[...]

PR #1273, opened 2021-06-16
batcher: don't crash on push failure, report error response

tommilligan:
See: #1152, #1150, #1027, #432, #375

To summarise: once a batch passes, bors will attempt to push the branch. If this fails, bors can crash. This is visible in the logs, but invisible in the relevant PRs:

image

This PR prevents bors from ever crashing silently as the result of a failed push. bors will instead mark the batch as errored, and push an appropriate message to the affected PRs.

This message will include the raw error response, which often contains further details.

As implemented here, bors will not retry or bisect the batch, as it is expected most failure cases are to do with a config/setup error outside the control of bors. I am open to changing this behaviour if there is strong feeling otherwise, but it is back-compatible in the sense that all current conditions that cause a crash do not trigger a retry of the batch. This PR has the same behaviour, but with a more visible error message.

### DIFF
--- a/lib/worker/batcher.ex
+++ b/lib/worker/batcher.ex
@@ -692,23 +692,22 @@ defmodule BorsNG.Worker.Batcher do
         batch.into_branch
       )
 
-    known_push_failure =
+    push_status =
       case push_result do
         {:error, :push, 422, raw_error_content} ->
-          String.contains?(raw_error_content, "Update is not a fast forward")
+          cond do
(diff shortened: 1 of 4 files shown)


notriddle (approved):
bors r+

This does seem like an improvement. Thanks!

tommilligan:
@notriddle thanks for reviewing and merging this so quickly!

I believe all the issues I linked in my initial description can now be closed as fixed. I will drop a comment in the relevant issues to make the original authors aware a fix is in master.

PR #1320, opened 2021-09-16
Improve batch details status readability

browdus:
The current batch details status list is simply a space delimited list
of all statuses. This can be quite difficult to read with more than one
or two statuses. Instead, use an unordered list, with one status per
line to improve readability with many statuses.

**Current**
current

**With Unordered List**
list

**Empty Unordered List**
empty


notriddle (approved):
bors r+

Issue #1239, opened 2021-05-05
bors gets stuck after a db crash


saifelse:
Hello! I am currently running bors on Build 524eff4. After `bors merge`ing on a PR, occasionally, I run into the following error:

```elixir
{%DBConnection.ConnectionError{message: "ssl recv: closed"},
 [
   {Ecto.Adapters.Postgres.Connection, :execute, 4,
    [
      file: 'lib/ecto/adapters/postgres/connection.ex',
      line: 108
    ]},
   {Ecto.Adapters.SQL, :sql_call, 6,
    [file: 'lib/ecto/adapters/sql.ex', line: 256]},
   {Ecto.Adapters.SQL, :execute_or_reset, 7,
    [file: 'lib/ecto/adapters/sql.ex', line: 436]},
   {Ecto.Repo.Queryable, :execute, 5,
    [file: 'lib/ecto/repo/queryable.ex', line: 133]},
   {Ecto.Repo.Queryable, :all, 4,
    [file: 'lib/ecto/repo/queryable.ex', line: 37]},
   {BorsNG.Worker.Batcher, :do_handle_cast, 2,
    [file: 'lib/worker/batcher.ex', line: 184]},
   {BorsNG.Worker.Batcher, :handle_cast, 2,
    [file: 'lib/worker/batcher.ex', line: 94]},
   {:gen_server, :try_dispatch, 4,
    [file: 'gen_server.erl', line: 637]}
 ]}
```
This is likely an issue with how we have our database set up, but it has a surprising behavior: after the crash, the PR ends up in a weird state:
 - In Bors UI, it's corresponding Batch is marked as "Canceled".
 - But in GitHub, the status is stuck in a `Running` state. Running `bors r+` /  `bors r-` end up having no impact. One way to un-stuck the PR is to push a commit, and then run `bors r+` again.

My hunch is that this could be related to #256 where there is no longer transactionality?, but I don't know the bors data model / codebase well enough.  Please let me know if there's any other information I can provide or if there's anything worth looking into the next time I run into this!

PR #1321, opened 2021-09-22
Allow disabling database prepared statements

saifelse:
Expose Ecto.Adapters.Postgres's `:prepare` [connection option](https://hexdocs.pm/ecto_sql/Ecto.Adapters.Postgres.html#module-connection-options):
via the env var `DATABASE_PREPARE_MODE`:
- `named` (default) preserves the existing behavior of using prepared statements
- `unnamed` prevents Ecto from using prepared statements

This was introduced in Postgrex v0.11.1 and so this config should already be available.

---

When running bors-ng with a Postgres database in Amazon's Aurora Serverless,
one runs into daily connection issues as described in #1239, as Aurora
Serverless automatically [closes connections that are older than 24 hours](
https://aws.amazon.com/blogs/database/best-practices-for-working-with-amazon-aurora-serverless/
):
> Aurora Serverless closes connections that are older than 24 hours. Make sure
> that your connection pool refreshes connections frequently.

Unfortunately, Ecto's connection pool can't be configured to refresh
connections frequently, so one must run a connection pooler between Ecto and the
Postgres database that does have this capability, like PgBouncer.

But, since Ecto itself tries to maintain a connection pool, it will typically
have long-lived sessions, preventing PgBouncer with its default configuration
from aggressively refreshing the connections. One promising mode of PgBouncer
is transaction pooling, which allows PgBouncer to clean up connections after
transactions end, but prevent's Ecto's default behavior of using prepared
statements. The `:prepare` config flag allows disabling this behavior, and was
added specifically for this purpose. See: elixir-ecto/postgrex#156

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,7 @@ All the same recommendations apply, with some extra notes:
 - `ALLOW_PRIVATE_REPOS` must be set at both build and run times to take effect. It is set to ` true` by default.
 - `DATABASE_URL` *must* contain the database port, as it will be used at container startup to wait until the database is reachable. [The format is documented here](https://hexdocs.pm/ecto/Ecto.Repo.html#module-urls).
 - `DATABASE_TIMEOUT` may be set higher than the default of `15_000`(ms). This may be necessary with repositories with a very large amount of members.
+- `DATABASE_PREPARE_MODE` can be set to to `unnamed` to disable prepared statements, [which is necessary when using a transaction/statement pooler, like pgbouncer](https://github.com/elixir-ecto/postgrex#pgbouncer). It is set to `named` by default.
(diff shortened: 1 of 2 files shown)


saifelse:
I was able to test this locally with pgbouncer and confirmed it met my use case... but it was a bit difficult to set up. Let me know if reproducible steps would be helpful.

notriddle (approved):
It's fine.

bors r+

dblackdblack (approved):

saifelse:
@notriddle : thanks for the swift review! Would it be possible to also get a release to docker hub?

notriddle:
@salfelse Okay, I've pushed a new Docker release (and since Docker is now charging for automated builds, I've also added a checkmark to run these releases at the start of every month, so that it doesn't fall out-of-date like this any more).

PR #1347, opened 2021-10-06
fix(github): respect API timeout at individual HTTP level

### LINKED ISSUE #1323: Bors crashing when trying to merge PRs.
We have bors running as Azure App service and our GitHub Ent. Server (GHES v3.1.6) instance with installed bors app. When we trying to bors r+ several PRs, bors reports Merge conflict status(not a message and there is no actual conflict) and we see the next error in log:
```
{:timeout,
 {GenServer, :call,
  [
    BorsNG.GitHub,
    {:merge_branch, {{:installation, 45}, 3038},
     {%{
        commit_message: "[ci skip][skip ci][skip netlify] -bors-staging-tmp-45",
        from: "0ed4ed02a910f3859910264d498890befe092e19",
        to: "staging.tmp"
      }}},
    10000
  ]}}
```
We've tested the same on GHES v3.1.4 (deployed on Azure VM) and there is no such issue.
What could be the problem?

notriddle:
It would be better to support deadlines, but I don't know how to get
that with the OTP GenServer receive timeouts.

Fixes #1323

### DIFF
--- a/lib/github/github/server.ex
+++ b/lib/github/github/server.ex
@@ -957,7 +957,9 @@ defmodule BorsNG.GitHub.Server do
 
     Tesla.client(
       middleware,
-      Tesla.Adapter.Hackney
+      {Tesla.Adapter.Hackney,
+       connect_timeout: Confex.get_env(:bors, :api_github_timeout, 8_000) - 1,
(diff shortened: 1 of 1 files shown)


notriddle:
Bors r+

PR #1387, opened 2021-11-11
Fix a typo

kidonng:
Hi, unless I missed something, requesting up-to-date branches "fixes the problem" so "problems like this can<strong>'t</strong> show up".


notriddle (approved):
bors r+

PR #1401, opened 2021-11-30
Ignoring review entries without commit_id

gfleury:
Github Enterprise latest versions has more review types, some of them do not have commit_id node, and of course this is causing it to fail.

The test added cover the types of reviews available.


notriddle (commented):
Awesome, thanks!

bors r+

notriddle (approved):

PR #1452, opened 2022-02-02
Load database repo dynamically

robertkowalski:
When creating an elixir release, the database adapter is hardcoded
into `sys.config`. With the hardcoded database, the docker image
is strictly set to Postgres.

This change allows to set the database url to a MySQL host by
environment variable, so the current docker image can connect to
MySQL and use the appropiate Ecto Adapter.

Co-authored-by: Björn Brauer <bjoern.brauer@new-work.se>
Co-authored-by: Theodor Fiedler <theodor.fiedler@new-work.se>
Co-authored-by: Phillip Hinrichsen <phillip.hinrichsen@new-work.se>
Co-authored-by: Markus Wolf <markus.wolf@new-work.se>


lib/database/repo.ex
```diff
@@ -1,3 +1,16 @@
+defmodule MethodMissing do
+  defmacro __using__(_opts) do
```
robertkowalski, line 2: dialyzer does not like it:

```
Starting Dialyzer
[
  check_plt: false,
  init_plt: '/home/user/repo/_build/dev/dialyxir_erlang-24.0.1_elixir-1.12.0_deps-dev.plt',
  files: ['/home/user/repo/_build/dev/lib/bors/ebin/Elixir.BorsNG.GitHub.Server.beam',
   '/home/user/repo/_build/dev/lib/bors/ebin/Elixir.BorsNG.ErrorView.beam',
   '/home/user/repo/_build/dev/lib/bors/ebin/Elixir.BorsNG.Database.AttemptState.beam',
   '/home/user/repo/_build/dev/lib/bors/ebin/Elixir.BorsNG.GitHub.FriendlyMock.beam',
   '/home/user/repo/_build/dev/lib/bors/ebin/Elixir.BorsNG.Router.Helpers.beam',
   ...],
  warnings: [:no_unused, :error_handling, :race_conditions, :unknown]
]
Total errors: 142, Skipped: 0, Unnecessary Skips: 0
done in 0m4.93s
lib/database/context/dashboard.ex:0:call_to_missing
Call to missing or private function BorsNG.Database.Repo.all/1.
________________________________________________________________________________
lib/database/context/dashboard.ex:0:call_to_missing
Call to missing or private function BorsNG.Database.Repo.all/1.
__________________________________________________________________
```

the alternative is to use `BorsNG.Application.fetch_repo()` for all `Repo.` calls in the repository

what do you think?

notriddle: Could we use wrapper methods? Perhaps generated by a macro?

```
def all(q): BorsNG.Application.fetch_repo().all(q)
def all(q, opts): BorsNG.Application.fetch_repo().all(q, opts)
def get(q, id, opts): BorsNG.Application.fetch_repo().get(q, id, opts)
def get!(q, id, opts): BorsNG.Application.fetch_repo().get(q, id, opts)
... etc
```

notriddle:
Also, thanks a bunch for this! I really do want bors-ng to be easy enough to deploy, and being able to use a database that you have operational competency in is great.

lib/database/repo.ex
robertkowalski: @notriddle we added a macro to define the functions dynamically from a file

lib/database/repo.ex
```diff
@@ -1,48 +1,55 @@
...
+  @ecto_repo_callbacks
+  |> Enum.flat_map(fn
+    {callback, arity} when arity >= 1 ->
+      is_defined = Enum.any?(@ecto_repo_callbacks, fn fun -> fun == {callback, arity - 1} end)
 
-    {:ok, config}
-  end
+      if is_defined,
+        do: [{callback, arity}],
+        else: [{callback, arity}, {callback, arity - 1}]
```
robertkowalski, line 45: we are not super excited about this way to define callbacks that take into account all arities. it would be much nicer if we could define the functions with default arguments in the first place, but we couldn't make that work with `Macro.var`.

Maybe you have another idea @notriddle?

notriddle (approved):
bors r+

I’m not really *thrilled* by it either, but I can live with it.

lib/application.ex
```diff
@@ -5,14 +5,33 @@ defmodule BorsNG.Application do
 
   use Application
 
+  def set_repo do
+    repo_module =
+      case System.get_env("BORS_DATABASE") || "postgresql" do
```
ucwaldo, line 10:
```suggestion
      case System.get_env("BORS_DATABASE", "postgresql") do
```

config/test.exs
```diff
@@ -1,8 +1,8 @@
 use Mix.Config
 
-case System.get_env("BORS_TEST_DATABASE") do
+case System.get_env("BORS_DATABASE") || "postgresql" do
```
ucwaldo, line 3: Is the case statement really needed here? I dont think it does much harm to always configure both repos

lib/database/repo.ex
```diff
@@ -1,48 +1,55 @@
...
+    []
+  end
 
-  def init(_, config) do
-    # Backwards compatibility hack: if POSTGRES_HOST is set, and the database URL is left at default,
-    # use the older configuration.
-    config = Confex.Resolver.resolve!(config)
-    no_host = is_nil(System.get_env("POSTGRES_HOST"))
+  def create_ecto_repo_callback_args(module, arity) do
+    Enum.map(1..arity, &Macro.var(:"arg#{&1}", module))
```
ucwaldo, line 19: Any way to define default arguments here, sparing us from having to define the functions explicitly with multiple arities?

config/prod.secret.exs
```diff
@@ -1,6 +1,6 @@
 use Mix.Config
 
-config :bors, BorsNG.Database.Repo,
+config :bors, BorsNG.Database.RepoPostgres,
```
ucwaldo, line 3: Is this project also released with MIX_ENV=prod? If yes, we should probably add a config for `BorsNG.Database.RepoMysql` here as well.

lib/mix/bors.cleanup.ex
```diff
@@ -67,5 +67,5 @@ defmodule Mix.Tasks.Bors.Cleanup do
     end
   end
 
-  def repos, do: Application.get_env(:bors, :ecto_repos, [])
+  def repos, do: [BorsNG.Application.fetch_repo()]
```
ucwaldo, line 70: Should we replace those with an import:

```
import BorsNG.Application, only: [fetch_repo: 0]

def repos, do: [fetch_repo()]
```

lib/database/repo.ex
```diff
@@ -1,48 +1,55 @@
...
-
-  Do not confuse this with a GitHub repo.
-  We call those `Project`s internally.
+  Exposes a macro to define Ecto.Repo functions dynamically.
   """
+  defmacro create_ecto_repo_callback(args, name) do
+    quote bind_quoted: [args: args, name: name] do
+      def unquote(name)(unquote_splicing(args)) do
+        repo = :persistent_term.get(:db_repo)
+        apply(repo, unquote(name), unquote(args))
```
ucwaldo, line 9: Not having much experience with macros, is `apply` really the only way to go here?

config/prod.secret.exs
notriddle: Yes, bors does get released with MIX_ENV=prod. You can see that in the docker file

https://github.com/bors-ng/bors-ng/blob/master/Dockerfile#L25

robertkowalski:
@notriddle we cleaned up the PR a bit, and retested the docker image. we are ready now and happy to squash :)

notriddle:
Getting a 403 in the logs. Updated perms, let's try it again.

robertkowalski:
```
Generated bors app
** (Mix) Could not find migrations directory "priv/repo_postgres/migrations"
for repo BorsNG.Database.RepoPostgres.
```

on it!

notriddle (approved):
bors r+

notriddle (approved):
bors r+

robertkowalski:
@notriddle we are running into an issue with `mix test` - the database is not set up properly. we introduced a manual ecto.create and migrate for both postgres and mysql before `mix test` but during `mix test` these tables seem to have disappeared.

https://github.com/xing/bors-ng/blob/a4053d3d0a7fc9a54c6ab41ff2561ccd1d91c873/config/config.exs#L24 and  https://github.com/xing/bors-ng/blob/a4053d3d0a7fc9a54c6ab41ff2561ccd1d91c873/priv/repo/migrations/20170823012140_change_users_login_type_to_citext.exs#L4-L17 are relevant lines of code

any ideas how we can fix this? passing `-r BorsNG.Database.RepoMysql` to `mix test` does not work, also passing `BORS_DATABASE=mysql` did not work, `mix test` seems to do its own thing...

robertkowalski:
@notriddle nevermind, we fixed it :) thanks for rubber ducking :)

robertkowalski:
great! thanks @notriddle! can you publish a new docker release so we can deploy on our side?

PR #1502, opened 2022-05-16
Read system environment variables for HTTPS proxy

### LINKED ISSUE #1500: Unable to deploy with squid proxy
Hello, we are using a squid proxy for our apps, and would like bors to use it when calling Github's API. I tried setting the usual HTTPS_PROXY or https_proxy environment variable on Linux and it doesn't seem to go through it. I'm not familiar with elixir or erlang and I wasn't able to find documentation on how to make it use this.

notriddle:
Fixes #1500

CC @jonathan-boudreau-work

### DIFF
--- a/lib/github/github/server.ex
+++ b/lib/github/github/server.ex
@@ -954,11 +954,22 @@ defmodule BorsNG.GitHub.Server do
         middleware
       end
 
+    params =
+      [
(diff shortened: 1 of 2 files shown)


jonathan-boudreau-work (approved):
Tested with squid and it works 👍  thanks for this

PR #1601, opened 2023-01-13
Make failure messaging a bit more friendly

rl-king:
Replaces 👎 with ⚠️  for failures and 👎 with 🕐 for a "waiting" message


rl-king:
This was supposed to be PR'ed for a fork

PR #1602, opened 2023-01-13
Set `Logger` level using `BORS_LOG_LEVEL` if provided

jali-clarke:
Sometimes when the Github API is on the fritz it can be useful to have debug-level logging.  Setting the log level in `config.exs` and rebuilding the container image is not ideal - having something more dynamic would be very nice to have.

This PR introduces the `BORS_LOG_LEVEL` environment variable which is converted to a `Logger` level atom using `String.to_existing_atom` when provided.  If `BORS_LOG_LEVEL` is unset, the `Logger` level will remain unchanged and is inherited from `config.exs` (i.e. `:info` for prod release).

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,7 +339,8 @@ Or you can do it manually:
         GITHUB_CLIENT_SECRET=<OAUTH_CLIENT_SECRET> \
         GITHUB_INTEGRATION_ID=<ISS> \
         GITHUB_INTEGRATION_PEM=`base64 -w0 priv.pem` \
-        GITHUB_WEBHOOK_SECRET=<SECRET2>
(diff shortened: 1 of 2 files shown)


jali-clarke:
@notriddle We have tested this and it does work - we're even able to get a console into the running process using `bors rpc` and setting `Logger.configure(level: :debug)`.  We are consequently able to see debug log messages being emitted.

That purging only happens [if `:compile_time_purge_matching` is used](https://hexdocs.pm/logger/1.14.2/Logger.html#module-application-configuration), and it's not being used.  I'll add a comment to that effect in the code.

jali-clarke:
Fixing formatting issues...

jali-clarke:
@notriddle Amended and force-pushed.

PR #1608, opened 2023-02-15
Move from deprecated `set-output` command to `GITHUB_OUTPUT` environment variable

tanguy-platsec:
To mitigate the following warning:
```
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
Indeed, according to https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/, workflows still using this keyword will start failing on 1st June, 2023, and this day will come soon enough.

Needs to be tested using the branch name as a tag in workflows that use it.


tanguy-platsec:
Sorry for closing it!
The commit history is a bit dirty though so I suggest you squash these when merging, sorry for that!
Have a great day :)

PR #1614, opened 2023-03-05
Initial steps towards implementing RFC 0322

A5rocks:
RFC: https://bors.tech/rfcs/0322-pre-test-and-pre-merge-hooks.html

I presume I don't have to open an issue for implementing one?

At the moment this only adds some (unpolished, incomplete) support for hooks only in the pre-test phase. I will finish this off soon but I would prefer to have some immediate feedback on whether anything I'm doing is conceptually wrong / there's a bug / there's a simpler approach. After all I'm very much missing the forest for the trees here and my mostly-forgotten knowledge of Elixir isn't helping me understand the architecture very much!

Thank you in advance for any time spent checking my code :P

### DIFF
--- a/lib/database/hook.ex
+++ b/lib/database/hook.ex
@@ -1,0 +1,56 @@
+defmodule BorsNG.Database.Hook do
+  @moduledoc """
+  A database record for an individual hook call.
+
+  See https://bors.tech/rfcs/0322-pre-test-and-pre-merge-hooks.html
+  """
+
+  @type t :: %__MODULE__{}
+  @type state_n :: 0 | 1 | 2 | 3
+  @type state :: :queued | :pending | :ok | :error
+
+  use BorsNG.Database.Model
+  alias BorsNG.Database.HookState
+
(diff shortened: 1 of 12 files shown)


A5rocks:
Yeah that sounds like what I did. I'm not entirely sure that it's a good idea to `setup_statuses` after the hooks are complete (because race conditions -- if no commit gets pushed and statuses finish before hooks do, then that status will fall through the cracks) but at the same time I'm not sure how exactly I should invalidate statuses when new commits get pushed.

A5rocks:
I was going to try to add in pre-try hooks but the duplication scared me off (and also the only hook kind I personally need are pre-test hooks :-)

However, this PR still completely implements the flow for pre-test hooks, I believe. As such, I think this is probably as far as I'll get!

A5rocks:
Hi, this is actually ready for a review I think! Not sure why the style check is failing.... (not at a computer right now but if I remember correctly I could not repro locally? I'll check again soon!)

PR #1670, opened 2023-06-01
Add deprecation notice to README

notriddle:


Gabriella439:
I think it's worth noting that there is at least one reason why GitHub's merge queue is not yet usable, which is this issue: https://github.com/orgs/community/discussions/36568

At least, that's the reason we've been pushing on getting `bors-ng` working recently instead of using GitHub's merge queue.  I think once GitHub's merge queue is out of beta and doesn't have issues like those then it'd make more sense to deprecate `bors-ng`

